PR #16, opened 2018-01-23
Modified setup.py to ensure that it works in 2.65, 2.7.5, & 3.4.5

MTWiley:
sys.version_info.major is not supported in all versions, but the alternative sys.version[0] appears to function in every version I've been able to test with.


Noname2244:
If this related ? Can't install pyntc

sudo python setup.py install
  File "setup.py", line 29, in <module>
    if sys.version_info.major >= 3:

MTWiley:
@Noname2244 Yep, that's exactly what this was written to address.

jmcgill298:
This is taken care of in develop branch

PR #19, opened 2018-04-13
added support for custom transport port on EOS

viniarck:
This enables users to use a custom transport port, if they are running transport `http` on a different port other than 80, or transport `https` on a different port other than 443.

Also, I've formatted the lines to have fewer than 80 characters.


itdependsnetworks:
It's failing since it is pointing to an older netmiko naming convention. Will update then kick this CI off again.

viniciusarcanjo:
Right. Thanks, @itdependsnetworks

PR #20, opened 2018-04-14
update netmiko import in tests

itdependsnetworks:


jmcgill298:
IOS Tests have been fixed

PR #21, opened 2018-04-17
Updated dependencies for pyeapi and textfsm

sheffner1226:
The pyeapi and textfsm modules are now natively supported in Python3.  Updated the dependencies to reflect that because the existing dependencies for pyeapi and textfsm listed versions that no longer exist.  The previous textfsm==9.9.9 was renamed to jtextfsm and is now version 0.3.1.  The textfsm module now supports Python3 and is on version 0.3.2.  I didn't specify versions for those modules because there doesn't appear to be a need.  If that is desired, though, the current versions are:  PyEAPI:  0.8.2 and TextFSM: 0.3.2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,8 +17,14 @@ install_requires = [
     'coverage',
     'mock>=1.3',
     'terminal',
-    'textfsm',
 ]
+
+if sys.version_info.major >= 3:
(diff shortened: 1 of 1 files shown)


jmcgill298:
This is fixed in develop branch, and will in master soon

PR #22, opened 2018-05-19
Fix issues with Junos mock since they switched to implementing metaclass

jmcgill298:


jmcgill298:
@itdependsnetworks I believe this fixes the Juniper Mock issue, but there are still other issues with netmiko

Issue #14, opened 2017-11-07
pip3 install pyntc


hosseinoliabak:
I am receiving the following error when installing pyntc for python 3. I have no issue with python 2. I tried 3 different Ubuntu Machines still the same message!
```
pip3 install pyntc
 Could not find a version that satisfies the requirement textfsm==1.0.1 (from pyntc) (from versions: 0.3.2)
No matching distribution found for textfsm==1.0.1 (from pyntc)
```
Please guide accordingly.
Regards,

richwalker1988:
I too am having the same issue

please help

lboue:
Hello,

I am facing the same issue on Ubuntu 17.10 too:
```
user@ubuntu:~$ uname -a
Linux ubuntu 4.13.0-16-generic #19-Ubuntu SMP Wed Oct 11 18:35:14 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

```
user@ubuntu:~$ sudo pip3 install pyntc         
[sudo] password for user: 
The directory '/home/user/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
The directory '/home/user/.cache/pip' or its parent directory is not owned by the current user and caching wheels has been disabled. check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
Collecting pyntc
  Downloading pyntc-0.0.5.tar.gz
Collecting coverage (from pyntc)
  Downloading coverage-4.4.2-cp36-cp36m-manylinux1_x86_64.whl (197kB)
    100% |████████████████████████████████| 204kB 69kB/s 
Collecting future (from pyntc)
  Downloading future-0.16.0.tar.gz (824kB)
    100% |████████████████████████████████| 829kB 236kB/s 
Collecting jsonschema (from pyntc)
  Downloading jsonschema-2.6.0-py2.py3-none-any.whl
Collecting mock>=1.3 (from pyntc)
  Downloading mock-2.0.0-py2.py3-none-any.whl (56kB)
    100% |████████████████████████████████| 61kB 64kB/s 
Collecting netmiko (from pyntc)
  Downloading netmiko-1.4.3.tar.gz (47kB)
    100% |████████████████████████████████| 51kB 122kB/s 
Collecting paramiko (from pyntc)
  Downloading paramiko-2.4.0-py2.py3-none-any.whl (192kB)
[...]

PR #36, opened 2018-09-12
Fix dependency

GGabriele:
Trying to address [#14](https://github.com/networktocode/pyntc/issues/14)
Not sure why `pyeapi==9.9.9` was there, since the latest version is actually `0.8.2`


TomCos:
Looks like it was in there to catch python3 being used, but the pyeapi package supports v3 now, so I don't see the issue.

GGabriele:
Definitely a fancy way to handle versions support :)

jmcgill298:
This is resolved in develop branch

PR #43, opened 2018-11-01
added a reconnect option

amb1s1:


pyntc/devices/ios_device.py
```diff
@@ -69,6 +69,18 @@ def _raw_version_data(self):
...
+        counter = 0
+        timeout = timeout*4
+        while counter < timeout:
+            try:
+                self.open()
+                return True
+            except:
+                counter += 1
+                time.sleep(15)
+        raise ValueError('reconnect timeout: could not verified device upgrade')
```
jmcgill298: The exception should be generic since it is not necessarily tied to upgrading the OS.

pyntc/devices/ios_device.py
```diff
@@ -69,6 +69,18 @@ def _raw_version_data(self):
             return version_data
         except IndexError:
             return {}
+            
+    def _reconnect(self, timeout=60):
+        counter = 0
+        timeout = timeout*4
```
jmcgill298: The timeout should be a second counter, and then use time library to measure length of time.

PR #50, opened 2018-11-08
Add install_os method to the Base Device Model

mzbroch:
Add install_os method to the Base Device Model

### DIFF
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -175,6 +175,21 @@ class BaseDevice(object):
         raise NotImplementedError
 
     @abc.abstractmethod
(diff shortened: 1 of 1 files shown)


pyntc/devices/base_device.py
```diff
@@ -174,6 +174,21 @@ def get_boot_options(self):
...
+
+        Args:
+            image_name(str): Image's Name
+
+        Returns:
+            True if system has been installed during function's call, False if OS has not been installed
+
+        Raises:
+            RuntimeError: If there is a problem with OS installation.
+        """
```
jmcgill298: We should add a list of acceptable args for `vendor_specifics`

PR #54, opened 2018-11-09
FILE COPY: Enhance functionality of file copy

jmcgill298:
* Use internal method, `_get_file_system`, to dynamically assign default file system

 * Update `file_copy` method to only copy files when the file does not exist on the remote device

 * Add validation that file was copyied correctly post file transfer


amb1s1:
LGTM

PR #61, opened 2018-11-13
Suggest to make F5 install_os method to be more inline with IOS & NXOS

jmcgill298:
* F5Device: change `wait_for_image_installed` to return `None` when successful and raise an exception when it fails.

 * F5Device: Move "free space" check for file upload to `file_copy` method so that all file copies perform this check.

* F5Device: Add call to `file_copy` method as part of the `install_os` method process (file_copy validates copy is missing before transferring).

* F5Device: remove exception raising in `install_os` method as the methods it depends on will raise exceptions for failures.


jmcgill298:
@mzbroch can you take a look at this and let me know what you think? The basic idea was to make this more like the other classes. The goal of the other classes was to have methods that performed an action validate their work, and raise an exception if they failed to accomplish it (instead of returning True/False).

PR #81, opened 2019-10-13
Pin junos-eznc

FragmentedPacket:
junos-eznc breaks f5-sdk package. Quick fix to pin junos-eznc to 2.2.1

See https://github.com/networktocode/ntc-ansible/issues/253 for details


jmcgill298:
@FragmentedPacket Probably should also update imports to have F5 imports happen before junos-eznc imports.

jmcgill298:
Develop has fixed this with moving to Poetry

PR #109, opened 2020-05-20
Fix call to enable mode on device

### LINKED ISSUE #107: Bug with IOS enable mode
Below code needs to be updated as this line raises an exception when not in enable mode:

https://github.com/networktocode/pyntc/blob/9640259b77816060de8ecfa62d626bc77e99d124/pyntc/devices/ios_device.py#L46

Should probably just call self.native.enable and forgo the check

jmcgill298:
- Deprecates the private method `._enable()` on ASA and iOS devices.
- Creates a new public method `.enable()` which encapsulates the old private method behavior.
- Updates internal calls to `_enable` to use the new public API.

### DIFF
--- a/pyntc/devices/asa_device.py
+++ b/pyntc/devices/asa_device.py
@@ -5,6 +5,8 @@ import os
 import re
 import signal
 import time
+import warnings
+
 
 from netmiko import ConnectHandler
 from netmiko import FileTransfer
@@ -40,7 +42,7 @@ class ASADevice(BaseDevice):
         self.open()
 
     def _enable(self):
-        print("This method will be deprecated; migrate to use the ``enable`` method")
+        warnings.warn("_enable() is deprecated; use enable().", DeprecationWarning)
         self.enable()
 
     def _enter_config(self):
--- a/pyntc/devices/asa_device.py
+++ b/pyntc/devices/asa_device.py
@@ -180,6 +180,11 @@ class ASADevice(BaseDevice):
         self.native.exit_config_mode()
 
     def enable(self):
+        """Ensure device is in enable mode.
+
+        Returns:
+            None: Device prompt is set to enable mode.
+        """
         # Netmiko reports enable and config mode as being enabled
         if not self.native.check_enable_mode():
             self.native.enable()
--- a/pyntc/devices/asa_device.py
+++ b/pyntc/devices/asa_device.py
@@ -40,12 +40,11 @@ class ASADevice(BaseDevice):
         self.open()
 
     def _enable(self):
-        self.native.exit_config_mode()
-        if not self.native.check_enable_mode():
-            self.native.enable()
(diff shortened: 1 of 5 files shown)


jmcgill298:
Fixes #107

pyntc/devices/asa_device.py
```diff
@@ -40,12 +40,11 @@ def __init__(self, host, username, password, secret="", port=22, **kwargs):
         self.open()
 
     def _enable(self):
-        self.native.exit_config_mode()
-        if not self.native.check_enable_mode():
-            self.native.enable()
+        print("This method will be deprecated; migrate to use the ``enable`` method")
```
shakefu, line 43: Nice! I like that this is wrapping the new public API while maintaining backwards-compatibility.

Preferable to use actual deprecation warnings here, though, so they can be controlled and filtered by the Application.

(Short summary with examples: https://www.programcreek.com/python/example/61499/warnings.DeprecationWarning)

pyntc/devices/asa_device.py
```diff
@@ -180,13 +179,21 @@ def config_list(self, commands):
                 raise CommandListError(entered_commands, command, e.cli_error_msg)
         self.native.exit_config_mode()
 
+    def enable(self):
```
shakefu, line 182: If you're happy and you Know It, write a docstring! :D

pyntc/devices/ios_device.py
```diff
@@ -42,12 +42,11 @@ def __init__(self, host, username, password, secret="", port=22, **kwargs):
         self.open()
 
     def _enable(self):
-        self.native.exit_config_mode()
```
shakefu, line 45: Since this exact API is duplicated across multiple devices, should it be abstracted to a Superclass?

pyntc/devices/ios_device.py
```diff
@@ -180,6 +179,14 @@ def config_list(self, commands):
                 raise CommandListError(entered_commands, command, e.cli_error_msg)
         self.native.exit_config_mode()
 
+    def enable(self):
+        # Netmiko reports enable and config mode as being enabled
```
shakefu, line 183: Same re: superclass, docstring.

pyntc/devices/ios_device.py
```diff
@@ -42,12 +42,11 @@ def __init__(self, host, username, password, secret="", port=22, **kwargs):
         self.open()
 
     def _enable(self):
-        self.native.exit_config_mode()
-        if not self.native.check_enable_mode():
-            self.native.enable()
+        print("This method will be deprecated; migrate to use the ``enable`` method")
```
shakefu, line 45: Same re: warnings.

pyntc/devices/ios_device.py
jmcgill298: I do want to revisit the design, but not at this time; I want to focus on making sure everything works and tests are working again.

PR #116, opened 2020-06-01
Pytest

shakefu:
- Adds flake8 and black to the `unit_test` 
- Adds flake8 config compatible with black


tox.ini
```diff
@@ -14,12 +15,12 @@ deps =
     coverage
 commands =
```
jmcgill298, line 16:
```suggestion
whitelist_externals = make
commands =
```

jmcgill298:
Thanks @shakefu glad to see green for a change

PR #118, opened 2020-06-03
Convert boot_options to a property

jmcgill298:


jmcgill298:
Travis logs show job kicking off and passing

PR #123, opened 2020-06-16
Start migration of utils into single directory

jmcgill298:
There are several directories that have files with little content, this starts migrating them into a single directory and simplifies the code.

Move textfsm templates into utils directory and update imports in ASA/IOS


pyntc/utils/__init__.py
```diff
@@ -0,0 +1 @@
+from .templates import get_structured_data  # noqa F401
```
shakefu, line 1: Same re: using `__all__` in lieu of `#noqa`

jmcgill298:
Github is not updating on the PR, will reopen

PR #131, opened 2020-07-12
Add Driver for WLC/AireOS

jmcgill298:
This is WIP in that I need to test a few methods against a real device, and will likely need to update some of the details of a couple of methods. The design, testing, and majority of the functionality should work.


petthomp:
looking to better understand how to create new Cisco nx-os template. I went through the process but looks like IM missing something. Do I need to request "pull" for the new template to work ?

Hoping you can help as I see you added to the group.

jmcgill298:
> looking to better understand how to create new Cisco nx-os template. I went through the process but looks like IM missing something. Do I need to request "pull" for the new template to work ?
> 
> Hoping you can help as I see you added to the group.

@petthomp I'm not sure what you are asking. What templates are you talking about? Did you mean to ask at [ntc-templates](https://github.com/networktocode/ntc-templates)?

What group are you referrencing?

jmcgill298:
I tested the reboot method today, and that will need some updates. I think I have the correct implementation, but need to test some more and update the test cases

test/unit/conftest.py
```diff
@@ -0,0 +1,141 @@
...
+
+@pytest.fixture
+def aireos_redundancy_state(aireos_redundancy_state_path):
+    with mock.patch(aireos_redundancy_state_path, new_callable=mock.PropertyMock) as rs:
+        rs.return_value = True
+        yield rs
+
+
+@pytest.fixture
+def aireos_redundancy_state_path(aireos_device_path):
```
mundruid, line 74: Not sure if we need fixtures for even returning paths, however I do like the fixture pattern.

test/unit/test_devices/test_aireos_device.py
```diff
@@ -0,0 +1,522 @@
...
+    convert_filename_to_version,
+)
+
+
+@pytest.mark.parametrize(
+    "filename,version",
+    (("AIR-CT5520-K9-8-8-125-0.aes", "8.8.125.0"), ("AIR-CT5520-8-8-125-0.aes", "8.8.125.0")),
+    ids=("encrypted", "unencrypted"),
+)
+def test_convert_filename_to_version(filename, version):
```
mundruid, line 21: Why not a TestAireosDevice class? I think this would be a better pattern and consistent with our other tests.

test/unit/test_devices/test_aireos_device.py
```diff
@@ -0,0 +1,522 @@
...
+
+def test_uptime_components(aireos_show):
+    device = aireos_show(["show_sysinfo.txt"])
+    days, hours, minutes = device._uptime_components()
+    assert days == 3
+    assert hours == 2
+    assert minutes == 20
+
+
+@mock.patch.object(AIREOSDevice, "open")
```
mundruid, line 71: If we had a test class we would be creating the mock in the setup method. That would be good for a next iteration.

PR #146, opened 2020-09-17
Add Support for AireOS AP Image Download

jmcgill298:


test/unit/test_devices/test_aireos_device.py
```diff
@@ -538,6 +640,137 @@ def test_show_list(mock_enable, aireos_send_command_timing):
...
+    mock_abo.assert_called_once()
+
+
+@mock.patch.object(AIREOSDevice, "config")
+@mock.patch.object(AIREOSDevice, "_wait_for_ap_image_download")
+@mock.patch.object(AIREOSDevice, "boot_options", new_callable=mock.PropertyMock)
+@mock.patch.object(AIREOSDevice, "_ap_images_match_expected")
+@mock.patch.object(AIREOSDevice, "ap_boot_options", new_callable=mock.PropertyMock)
+def test_transfer_image_to_ap_already_transferred_primary(
+    mock_abo, mock_aime, mock_bo, mock_wait, mock_config, aireos_device, aireos_mock_path, aireos_boot_image
```
mundruid, line 658: I suggest a little longer names that correspond to mock_boot_options, mock_image_match, etc.

test/unit/test_devices/test_aireos_device.py
```diff
@@ -538,6 +640,137 @@ def test_show_list(mock_enable, aireos_send_command_timing):
...
+@mock.patch.object(AIREOSDevice, "config")
+@mock.patch.object(AIREOSDevice, "_wait_for_ap_image_download")
+@mock.patch.object(AIREOSDevice, "boot_options", new_callable=mock.PropertyMock)
+@mock.patch.object(AIREOSDevice, "_ap_images_match_expected")
+@mock.patch.object(AIREOSDevice, "ap_boot_options", new_callable=mock.PropertyMock)
+def test_transfer_image_to_ap_transfer_secondary_fail(
+    mock_abo, mock_aime, mock_bo, mock_wait, mock_config, aireos_device, aireos_mock_path, aireos_boot_image
+):
+    mock_bo.return_value = {"primary": None, "backup": aireos_boot_image}
+    mock_aime.side_effect = [False, False, False, True, False]
```
mundruid, line 745: How about other combinations of these function return values, such as False, False, True, False, True? How about the first two arguments being true?

mundruid (commented):
Added two small editorial comments for your consideration. Thanks!

test/unit/test_devices/test_aireos_device.py
jmcgill298: This is a little tricky; if the 2nd response is True, then one of the responses is skipped (python returns early when the result of the conditional is determined, so True exists the `or` condition after the first test, and the second test never happens). The `[False, False, True, False, True]` is the same test as `[False, True, False, True]`, which is tested in `test_transfer_image_to_ap_already_transferred_primary`

[True, True] is not possible since the first true triggers the return value. This behavior is tested in `test_transfer_image_to_ap_already_active`

test/unit/test_devices/test_aireos_device.py
jmcgill298: I have updated all of the shortened names

jmcgill298:
Thanks for reviewing @mundruid

Issue #102, opened 2020-05-13
Change EOS file transfer to use Netmiko


jmcgill298:
We should be able to simplify file transfer code by using Netmiko, and then get rid of the pyntc/system_features/file_copy directory

PR #147, opened 2020-09-20
Change Eos Device to use Netmiko for File Transfer

mundruid:
- Addresses issue #102 
- Add Netmiko FileTransfer and ssh property
- Add corresponding tests, send_command


pyntc/devices/eos_device.py
```diff
@@ -46,6 +47,18 @@ def __init__(self, host, username, password, transport="http", timeout=60, **kwa
         self.timeout = timeout
         self.connection = eos_connect(transport, host=host, username=username, password=password, timeout=timeout)
         self.native = EOSNative(self.connection)
+        # _connected indicates Netmiko ssh connection
+        self._connected = False
+        # ftp corresponds to file transfer and openning a netmiko connection
+        if self.transport == "ftp":
```
jmcgill298, line 53: I don't think we would expect transport to be ftp. They will have an http/https transport, but if they use the file copy method, then we would establish an scp connection.

pyntc/devices/eos_device.py
```diff
@@ -46,6 +47,18 @@ def __init__(self, host, username, password, transport="http", timeout=60, **kwa
...
+            self.open()
```
jmcgill298, line 54: should wait to call `self.open()` until file copy method is called

jmcgill298 (changes requested):
Should also remove the existing file and directories for custom EOS file copy

jmcgill298:
Fixes #102

pyntc/devices/eos_device.py
mundruid: It was an unfortunate name. I wanted to distinguish between file copy and all other functions so that we do not create a netmiko ssh connection when we do not need it. Changed to address comment.

pyntc/devices/eos_device.py
mundruid: We could do open by default, however when we do not transfer files we do not need to have a netmiko SSH connection. It seemed like a waste, but it does not change the behavior to get rid of the if statement.

pyntc/devices/eos_device.py
mundruid: Changes made per our conversation

jmcgill298:
Thanks @mundruid

PR #149, opened 2020-10-03
Issue 136: modify send_command on all devices 

mundruid:
- Modify send_command to use expect_string. Only applies to devices: aireos, asa, and ios.
- Delete send_command_expect calls
- Change expect_string default value to None
- Modify unit tests to run with modified send_command not send_command_expect

### DIFF
--- a/test/unit/test_devices/test_aireos_device.py
+++ b/test/unit/test_devices/test_aireos_device.py
@@ -80,9 +80,9 @@ def test_send_command_timing(aireos_send_command_timing):
 def test_send_command_expect(aireos_send_command_expect):
     command = "send_command_expect"
     device = aireos_send_command_expect([f"{command}.txt"])
-    device._send_command(command, True, expect_string="Continue?")
-    device.native.send_command_expect.assert_called()
-    device.native.send_command_expect.assert_called_with("send_command_expect", expect_string="Continue?")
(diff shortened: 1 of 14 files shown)


test/unit/test_devices/test_aireos_device.py
```diff
@@ -75,14 +75,6 @@ def test_send_command_timing(aireos_send_command_timing):
     device.native.send_command_timing.assert_called_with(command)
 
 
-def test_send_command_expect(aireos_send_command_expect):
```
jmcgill298, line 78: I think we need to keep these tests, but change the assertion to `device.native.send_command.assert_called_with($args)`. This should be for all of them.

jmcgill298 (changes requested):
Overall it looks really good. The only thing is that I believe we should keep the existing test cases for expect, but modify how the tests are handled. Please let me know if you have any questions, or if my suggestion isn't working as expected.

test/unit/test_devices/test_aireos_device.py
mundruid: I have added back the test and changed the assertions.

jmcgill298 (approved):

jmcgill298:
Thanks @mundruid

PR #157, opened 2020-11-09
IOS: add properties/methods for redundancy info

jmcgill298:


pyntc/devices/ios_device.py
```diff
@@ -341,6 +395,56 @@ def handler(signum, frame):
         else:
             print("Need to confirm reboot with confirm=True")
 
+    @property
+    def redundancy_mode(self):
+        """
+        The oprating redundancy mode of the device.
```
mundruid, line 401: Change "oprating" to "operating"

test/unit/test_devices/test_ios_device.py
```diff
@@ -348,12 +348,121 @@ def test_enable_from_config(self):
...
+@pytest.mark.parametrize(
+    "redundancy_state,expected",
+    (
+        ("active", True),
+        ("standby hot", False),
+        (None, True),
+    ),
+    ids=("active", "standby_hot", "unsupported"),
+)
+def test_is_active(mock_redundancy_state, redundancy_state, expected, ios_device):
```
mundruid, line 361: Should there be a `test_is_not_active` for the `False` return or would this be too trivial?

mundruid (changes requested):

test/unit/test_devices/test_ios_device.py
jmcgill298: @mundruid that is covered by the second test case in the parametrize decorator.

pyntc/devices/ios_device.py
jmcgill298, line 401:
```suggestion
        The operating redundancy mode of the device.
```

jmcgill298:
Thanks for the review @mundruid

PR #158, opened 2020-11-10
Change IOS to verify device is active during open call

jmcgill298:


mundruid (approved):
Looks really good. Thank you for changing the tests to fixtures. I will be following this pattern from now on.

Issue #141, opened 2020-07-29
Increase Test Coverage


jmcgill298:
Coveralls reports test coverage at [82%](https://coveralls.io/github/networktocode/pyntc); we should start adding test cases for the files that have lower coverage percent

PR #161, opened 2020-11-13
Make AireOS FileTransferErrors more granular

jmcgill298:


mundruid (approved):
Looks good. From a quick look these new imports (mainly `FileTransferError`) will need to be tested for devices other than aireos. Do you want me to open an issue or do you think this is included in issue  #141: Increase test coverage?

jmcgill298:
This should be part of #141

PR #163, opened 2020-11-14
Allow for more time for aireos peer to form

jmcgill298:


pyntc/devices/aireos_device.py
```diff
@@ -277,6 +278,36 @@ def _wait_for_device_reboot(self, timeout=3600):
         # TODO: Get proper hostname parameter
         raise RebootTimeoutError(hostname=self.host, wait_time=timeout)
 
+    def _wait_for_peer_to_form(self, redundancy_state, timeout=300):
+        """
+        Wait for device redundancy state to form properly.
+
+        Args:
+            redundancy_state: The desired redundancy state between the system and its peer.
```
mundruid, line 286: It would be good to standardize new docstrings and add docstrings to "old" devices (issue #164 ). I would add type to the args, ex. redundancy_state (str) etc.

I use the Python docstring plugin on VS code to autogenerate these.

mundruid (changes requested):
Looks good. Just a small request in the effort to standardize docstrings. I also opened an issue #164  for these and I will be happy to add them. I have installed a plugin to generate these on VS code. :)

jmcgill298:
Thanks for the review @mundruid I fixed the errant doc strings and added them to all of the classes in `errors.py`

PR #166, opened 2020-11-23
Added bandit to tox test

qduk:
I also went through the code and put exceptions where they seemed applicable. I told bandit to skip [B101](https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html) as it didn't apply to us.


.bandit.yml
```diff
@@ -0,0 +1,2 @@
+---
+skips: ["B101"]
```
jmcgill298, line 2:
```suggestion
skips: ["B101"]

```

pyntc/devices/aireos_device.py
```diff
@@ -63,7 +63,7 @@ class AIREOSDevice(BaseDevice):
 
     vendor = "cisco"
 
-    def __init__(self, host, username, password, secret="", port=22, **kwargs):
+    def __init__(self, host, username, password, secret="", port=22, **kwargs):  # nosec
```
jmcgill298, line 66: what is the security issue being reported?

tox.ini
```diff
@@ -46,4 +51,4 @@ exclude =
     .pytest_cache
 
 # line length is handled by black
-ignore = E501
+ignore = E501
```
jmcgill298, line 54:
```suggestion
ignore = E501

```

pyntc/devices/aireos_device.py
jmcgill298: I think this is B107, which we should just ignore in the bandit config file

.bandit.yml
qduk: Updated.

pyntc/devices/aireos_device.py
qduk: That is correct. Added B107 to exclude list on bandit config file.

tox.ini
qduk: Updated

pyntc/devices/aireos_device.py
jmcgill298: Sorry, I meant that when adding that to the exclude list, you would remove the `# nosec` comments from the corresponding function defs

pyntc/devices/f5_device.py
```diff
@@ -285,7 +285,9 @@ def _upload_image(self, image_filepath):
                     end = size
                 content_range = "{}-{}/{}".format(start, end - 1, size)
                 headers["Content-Range"] = content_range
-                requests.post(_URI, auth=(self.username, self.password), data=payload, headers=headers, verify=False)
+                requests.post(
+                    _URI, auth=(self.username, self.password), data=payload, headers=headers, verify=False  # nosec
+                )
```
qduk, line 290: This is bandit test B501.

pyntc/devices/aireos_device.py
qduk: I did that. The function with the `nosec` still is bandit test B501.

jmcgill298 (approved):

PR #169, opened 2020-11-24
Adds Loop to the Dir 

jvanderaa:
Adds loop to attempt to get the dir output 5 times before failing to detect OS

### DIFF
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -77,13 +77,21 @@ class IOSDevice(BaseDevice):
         Raises:
             FileSystemNotFound: When the module is unable to determine the default file system.
         """
-        raw_data = self.show("dir")
-        try:
-            file_system = re.match(r"\s*.*?(\S+:)", raw_data).group(1)
-        except AttributeError:
-            raise FileSystemNotFoundError(hostname=self.facts.get("hostname"), command="dir")
+        # Set variables to control while loop
+        counter = 0
 
-        return file_system
+        # Attempt to gather file system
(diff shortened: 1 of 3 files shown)


jmcgill298 (approved):

pyntc/devices/ios_device.py
```diff
@@ -77,13 +77,21 @@ def _get_file_system(self):
...
-        raw_data = self.show("dir")
-        try:
-            file_system = re.match(r"\s*.*?(\S+:)", raw_data).group(1)
-        except AttributeError:
-            raise FileSystemNotFoundError(hostname=self.facts.get("hostname"), command="dir")
+        # Set variables to control while loop
+        counter = 0
+
+        # Attempt to gather file system
+        while counter < 5:
```
mundruid, line 84: Why specifically check 5 times? Can you please variabilize this with a constant to be able to change if needed?

mundruid (changes requested):
Hi Josh, the changes look good but the pipeline is failing black. Can you please fix the issues? Thanks!

pyntc/devices/ios_device.py
jvanderaa: Made into a variable/constant. 5 times as a number. Could maybe be 3. In the test it only worked 1 out of the 4 times. The other 3 had the error occur.

On the format front, not sure why the autoformat of VSCode would have different black settings than the project. I'm going to look into that yet.

jmcgill298:
Thanks @jvanderaa 

Thanks @mundruid for keeping us honest :)

Issue #99, opened 2020-05-13
Update facts attribute


jmcgill298:
The `facts` attribute should be converted to individual attributes and removed. Each attribute should be converted into a property or descriptor, and consideration should be given to how the attribute has information updated.

mundruid:
Addressed by PR #173

PR #172, opened 2020-11-28
Update facts attribute

mundruid:
- Broke down facts to individual properties in base device and all devices that had implemented facts
- Modified unit tests for properties
- Addresses issue #99


mundruid:
Had to create a new branch to fix conflicts with develop and issue with new facts implementation.

PR #175, opened 2020-12-02
Change IOS config methods to use native methods

jmcgill298:


jmcgill298:
@jvanderaa I was able to get IOS converted over

jvanderaa (approved):
Looks Good

pyntc/devices/ios_device.py
```diff
@@ -227,21 +254,93 @@ def close(self):
...
+        netmiko_args["config_commands"] = command
+        try:
+            command_response = self.native.send_config_set(**netmiko_args)
+        except TypeError as err:
+            raise TypeError(f"Netmiko Driver's {err.args[0]}")
+
+        self._check_command_output_for_errors(command, command_response)
+        return command_response
+
+    def config_list(self, commands, **netmiko_args):
```
mundruid, line 296: Doesn't `config_list` overlap with `config` since they can both accept a list of commands?

pyntc/devices/ios_device.py
```diff
@@ -227,21 +254,93 @@ def close(self):
...
+        """
+        Send config commands to device.
+
+        By default, entering and exiting config mode is handled automatically.
+        To disable entering and exiting config mode, pass `enter_config_mode` and `exit_config_mode` in ``**netmiko_args``.
+        This supports all arguments supported by Netmiko's `send_command_set` method using ``netmiko_args``.
+        This will send all commands in ``command`` regardless of failures, and only raise an Error after sending all commands.
+
+        Args:
+            command (str|list): The command or commands to send to the device.
```
mundruid, line 267: Seems that config should be able to handle a list of commands but the implementation does not indicate that.

pyntc/devices/ios_device.py
```diff
@@ -582,7 +681,7 @@ def set_boot_options(self, image_name, **vendor_specifics):
         try:
             command = "boot system {0}/{1}".format(file_system, image_name)
             self.config_list(["no boot system", command])
-        except CommandError:
+        except CommandListError:
```
mundruid, line 684: Why two different errors, one for command and one for commandList?

mundruid (commented):
I think it looks really good and the unit test rewrite is excellent. My only comment is: could we merge `config` with `config_list` they seem to be overlapping, but I may be missing something.

pyntc/devices/ios_device.py
jmcgill298: netmiko's `send_config_set` method supports both a single command as a string, out a list of commands to send. By switching to use the `send_config_set` method, we would have to do data validation, and I don't see what we would gain by enforcing a single command (open to hear the case for enforcement though)

pyntc/devices/ios_device.py
jmcgill298: CommandListError gives additional information explaining what command were sent before a failure was incurred.

Add for this update, this is actually the correct implementation for the existing code, as `config_list` raises this exception. It might still work, since `CommandListError` is raised when `CommandError` is raised.

pyntc/devices/ios_device.py
jmcgill298: There is overlap in that both accept sending a list of commands, but they did in how exceptions are handled, and the return date..

| config method | config_list method |
|-|-|
| All commands are sent before checking for an error | Commands are sent one at a time. Each command's result is checked for an error, and sending commands ceases upon detecting an error |
| Raises CommandError upon detecting an error, which would not be as helpful in determining which commands failed | Raises CommandListError upon detecting an error, which provides the list of commands that were sent, as well as which command failed |
| A string of all output from the device interaction is returned | A list of each command's interaction with the device is returned |

jmcgill298: I think if we want to consolidate these to a single interface, then we would need to do that across all OS Drivers. I'm not opposed to doing that, but would rather make that an issue to be addressed by other work.

pyntc/devices/ios_device.py
mundruid: I agree with all that you said. The only gain would be "do not repeat yourself" type of gain, but in this case it is small and I can really see the case for the different errors that the two commands are throwing.

pyntc/devices/ios_device.py
mundruid: Got it. Thanks for the response @jmcgill298

mundruid (approved):
Thank you for addressing my comment. I think we do not need to merge the config and config_list based on our conversation.

jmcgill298:
I have actually been thinking about your comment from a different vantage point, and have started to rework that to deprecate config_list, but still support both for now.

jmcgill298:
@mundruid I made the updates that I discussed this afternoon, how does this look?

mundruid (approved):
The new pattern with config looks good. Thank you for changing it @jmcgill298 .

PR #178, opened 2020-12-15
Add install mode ios

jvanderaa:
Creates an option to use Cisco's latest IOS upgrade feature, Install Mode. This is primarily for newer IOS models in the Catalyst line. This install mode will handle pushing the code to all members of an IOS stack, and reboot.  

There are two specific versions of code that requires a different command and this is accounted for with a brief search.  

From documentation at this time, the boot system file name **must** be `packages.conf`. This is a "constant" in the top of the code to handle this.


.bandit.yml
```diff
@@ -1,5 +1,5 @@
 ---
-skips: ["B101", "B107"]
+skips: ["B101", "B107", "B106"]
```
mundruid, line 2: B106: hardcoded_password_funcarg is an error that I do not think we should omit checking with bandit. Can't we add an exception for the directory / file if it is a test file? Please see lines 4-5 below. We do not want people to start... forgetting... hard coded passwords in the code base.

pyntc/devices/ios_device.py
```diff
@@ -541,12 +543,53 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
-            self.reboot(confirm=True)
+            if install_mode:
+                # Change boot statement to be boot system <flash>:packages.conf
+                self.set_boot_options(INSTALL_MODE_FILE_NAME, **vendor_specifics)
+                delay_factor = 10
+
+                # Check for OS Version specific upgrade path
+                # https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst9300/software/release/17-2/release_notes/ol-17-2-9300.html
+                os_version = self.os_version
+                if "16.5.1a" in os_version or "16.6.1" in os_version:
```
mundruid, line 569: Do we want to hard code OS versions? What if they change / become obsolete? Maybe add a constant at the beginning that can be changed?

pyntc/devices/ios_device.py
```diff
@@ -541,12 +543,53 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+            bool: False if no install is needed, true if the install completes successfully
+        """
         timeout = vendor_specifics.get("timeout", 3600)
         if not self._image_booted(image_name):
-            self.set_boot_options(image_name, **vendor_specifics)
-            self.reboot(confirm=True)
+            if install_mode:
+                # Change boot statement to be boot system <flash>:packages.conf
+                self.set_boot_options(INSTALL_MODE_FILE_NAME, **vendor_specifics)
+                delay_factor = 10
```
mundruid, line 564: Could the delay factor be included as another constant or even a positional argument? It may add some flexibility in fine tuning it.

test/unit/test_devices/test_ios_device.py
```diff
@@ -898,3 +898,81 @@ def test_set_boot_options_bad_boot(mock_save, mock_config, mock_boot_options, io
...
+    mock_set_boot_options,
+    mock_image_booted,
+    mock_os_version,
+    ios_device,
+):
+    image_name = "cat9k_iosxe.16.12.04.SPA.bin"
+    file_system = "flash:"
+    mock_get_file_system.return_value = file_system
+    mock_os_version.return_value = "16.6.1"
+    mock_image_booted.side_effect = [False, True]
```
mundruid, line 964: Maybe add some tests for all combinations: [False, False], [True, False], [True, True]?

test/unit/test_devices/test_ios_device.py
```diff
@@ -898,3 +898,81 @@ def test_set_boot_options_bad_boot(mock_save, mock_config, mock_boot_options, io
...
+    # Call the install_os
+    actual = ios_device.install_os(image_name, install_mode=True)
```
mundruid, line 966: Maybe add a test for install mode False?

mundruid (commented):
Josh, 

This feature looks good. I just had a couple of improvements / requests. Let me know if you would like to discuss these.

Thanks,
Xenia

.bandit.yml
jvanderaa: I'll test again, but there was a test that was failing on the EOS.

pyntc/devices/ios_device.py
jvanderaa: 👍

pyntc/devices/ios_device.py
jvanderaa: These are actually quite old versions at this point that we have moved on from. Current versions are 17.1 & 17.2, with a whole bunch of other dot versions 16.5, 16.6, 16.7, 16.8, 9, and more in between. I was thinking originally not including this at all, but it makes sense to cover our bases there.

Do you have any recommended methods of how to search list items in a string? The string inside of a list is easy. This one I'm not sure on.

test/unit/test_devices/test_ios_device.py
jvanderaa: `install_mode=false` is covered on the original install_os test (and was updated).

test/unit/test_devices/test_ios_device.py
jvanderaa: [True, False] is covered by `mock_image_booted.assert_called_once()` in the [True, True] test. Whenever the first mock is true it doesn't matter what the second one is as the boot should only be called once. (added in the second update)

.bandit.yml
jvanderaa: Removed from the exceptions.

pyntc/devices/ios_device.py
jvanderaa: If we want to stick on this but the tests are all passing, can we add an issue to update later?

mundruid: I am not sure if this is the most efficient way but I think you could make a list of os_versions then maybe check if the current os version is in this list:
```
list_os_versions = ['17.1', '17.2', '16.5', '16.7', '16.8', '16.9']
if os_version in list_os_versions:
    ...
```
Please let me know if I did not understand what you were trying to achieve.

jmcgill298: @mundruid I started to suggest something similar to this yesterday, but the lookup needs to happen in revers order, since `os_version` will be more than just the "16.x" portion. Whatever the final state of this PR is, we will want to go back later and refactor to handle install vs bundle modes better

jmcgill298 (approved):
LGTM

pyntc/devices/ios_device.py
mundruid: Got it @jmcgill298 , thanks for clarifying. Then it looks good to me as well!

mundruid (approved):

jmcgill298:
Thanks @jvanderaa

PR #179, opened 2020-12-16
Send command

jmcgill298:


jvanderaa:
Do we (@jmcgill298 or myself) need to update the ios_device to match the kwargs input? I believe in the IOS device we are just using `kwargs`, where as I think the Aireos is using `netmiko_kwargs`.

jmcgill298:
That will be handled in the next release when we add the netmiko base class

jvanderaa (approved):
LGTM

PR #180, opened 2020-12-16
Add deprecation to facts

jmcgill298:


jvanderaa:
Do we need a timeline for the deprecation warning? Otherwise makes sense based on what we worked on yesterday.

jmcgill298:
Deprecation will come with 1.0.0 release, which we don't have a timeline on. Since this is will be removed in the first "stable" release, I don't think we need to provide a timeline

jvanderaa (commented):
LGTM

PR #182, opened 2020-12-20
ASA add redundancy properties

jmcgill298:


pyntc/devices/asa_device.py
```diff
@@ -25,11 +25,16 @@
...
+RE_SHOW_FAILOVER_GROUPS = re.compile(r"Group\s+\d+\s+State:\s+(.+?)\s*$", re.M)
+RE_SHOW_FAILOVER_STATE = re.compile(r"(?:Primary|Secondary)\s+-\s+(.+?)\s*$", re.M)
+
+
 @fix_docs
 class ASADevice(BaseDevice):
     """Cisco ASA Device Implementation."""
 
     vendor = "cisco"
+    active_redundancy_states = {None, "active"}
```
mundruid, line 37: Why not make this data structure a list?

test/unit/test_devices/test_asa_device.py
```diff
@@ -304,6 +306,75 @@ def test_count_teardown(self):
...
+        (asa_module.CommandError("show failover", r"% invalid command"), None),
+    ),
+    ids=(ACTIVE, "standby", "disabled", "active_active", "active_standby", "standby_active", "none"),
+)
+def test_redundancy_state(side_effect, expected, asa_show):
+    device = asa_show([side_effect])
+    actual = device.redundancy_state
+    assert actual == expected
+
+
```
mundruid, line 377: The tests look great and I think they cover all possible combinations of states.

mundruid (approved):
I have one question: why did you use a dictionary for the set of active redundancy states?  Otherwise it looks good!

pyntc/devices/asa_device.py
jvanderaa: I like the usage of the set.

jvanderaa (approved):
LG

pyntc/devices/asa_device.py
jmcgill298: @mundruid The purpose of the data structure is to do lookups (is x in y?), which performs better in sets than lists

PR #186, opened 2020-12-21
Netmiko base

jmcgill298:


pyntc/devices/base_netmiko.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(
+        self,
+        host: str,
+        username: str,
+        password: str,
+        device_type: str,
+        secret: str = "",
+        port: int = 22,
+        **netmiko_args: Any,
```
mundruid, line 21: Why not include `global_delay_factor` as a property? It seems that it is used by ConnectHandler. It can be part of `**netmiko_args`, I just find it more clear when a property is common to use the specific property.

mundruid (approved):
I only had a comment for an additional property.

pyntc/devices/base_netmiko.py
jmcgill298: I would prefer not to define netmiko's optional or keyword args as changes in netmiko's calling signature do not affect us. For example, if Netmiko deprecated `global_delay_factor`, we would have to adjust our calling signature depending on what version of Netmiko the user had installed.

jvanderaa (approved):
LG

pyntc/devices/base_netmiko.py
mundruid: Makes sense, @jmcgill298. Thanks for the response!

PR #187, opened 2020-12-23
Add connected property to netmiko base

jmcgill298:


test/unit/conftest.py
```diff
@@ -320,3 +321,13 @@ def _mock(side_effects, existing_device=None, device=ios_device):
...
+
+
+@pytest.fixture
+def netmiko_device():
+    # with mock.patch.object(BaseNetmikoDevice, "confirm_is_active") as mock_confirm:
+    #     mock_confirm.return_value = True
+    with mock.patch("netmiko.ConnectHandler") as ch:
+        device = BaseNetmikoDevice("host", "user", "password", "cisco_ios_ssh")
+        device.native = ch
+        yield device
```
jvanderaa, line 333: I'm not seeing why yield here with no changes to what is being yielded. Should this just be a return instead?

jmcgill298: good catch, I have updated

jvanderaa (approved):
LGTM

PR #190, opened 2021-01-01
ASA add reload of standby device

jmcgill298:


pyntc/devices/asa_device.py
```diff
@@ -155,6 +156,45 @@ def _wait_for_device_reboot(self, timeout=3600):
...
+        """
+        start = time.time()
+        while time.time() - start < timeout:
+            if self.peer_redundancy_state == "failed":
+                break
+
+        while time.time() - start < timeout:
+            if self.peer_redundancy_state in acceptable_states:
+                return
+            time.sleep(1)
```
jvanderaa, line 193: Can this be put into a single while loop? If I read this right, it will start the check, only leave the while loop if it has a state of break. Then it will continue on, but since the time has elapsed it will only enter the RebootTimeOut Error. Never really reaching the second loop. 

I forget if the failed state will appear right away or not. Is that the purpose of the first while loop?

jmcgill298: The point of this is that the peer device will be dead before the connected device realizes it. Without the first while loop, the second loop will hit the return before the connected device realizes the peer is dead, which gives a false positive. The first while loop waits for the connected device to acknowledge failure of the peer device, then the second loop waits for the peer to come back online.

jvanderaa (approved):
LGTM

Issue #94, opened 2020-05-13
Device Reboot should not require confirm keyword


jmcgill298:
currently the interface for rebooting the device requires passing a confirm argument. This method should just reboot the device, and if a user does not want to reboot the device, then I suggest not using the method

PR #191, opened 2021-01-06
Remove confirm keyword requirement from device reboot

mundruid:
- Remove confirm keyword from all devices reboot
- Addresses issue #94

### DIFF
--- a/pyntc/devices/asa_device.py
+++ b/pyntc/devices/asa_device.py
@@ -379,7 +379,7 @@ class ASADevice(BaseDevice):
 
         return peer_redundancy_state.lower()
 
-    def reboot(self, timer=0):
+    def reboot(self, timer=0, **kwargs):
         """
         Reload the controller or controller pair.
 
@@ -394,6 +394,8 @@ class ASADevice(BaseDevice):
             >>> device.reboot()
             >>>
         """
+        if kwargs.get("confirm"):
+            warnings.warn("Passing 'confirm' to reboot method is deprecated.", DeprecationWarning)
 
         def handler(signum, frame):
             raise RebootSignal("Interrupting after reload")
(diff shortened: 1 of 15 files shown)


jmcgill298 (changes requested):
Thanks @mundruid, one comment, can you add `**kwargs` to the reboot method and send a warning that passing `confirm` is deprecated if confirm is in kwargs?

```
def reboot(..., **kwargs):
    if confirm in reboot:
        warnings.warn("passing 'confirm' to reboot method is deprecated..", DeprecationWarning)
```

mundruid:
> Thanks @mundruid, one comment, can you add `**kwargs` to the reboot method and send a warning that passing `confirm` is deprecated if confirm is in kwargs?
> 
> ```
> def reboot(..., **kwargs):
>     if confirm in reboot:
>         warnings.warn("passing 'confirm' to reboot method is deprecated..", DeprecationWarning)
> ```

Thanks for the review @jmcgill298 . It's a good point for backward compatibility and I added this. 

I added tests in the unit tests to check if the warning is triggered. We can remove them if we consider them unnecessary. I only added unit tests for two types unit with fixtures and without, i.e., asa and aireos, since the deprecation code is the same for all.

pyntc/devices/asa_device.py
```diff
@@ -379,31 +379,44 @@ def peer_redundancy_state(self):
...
 
-            def handler(signum, frame):
-                raise RebootSignal("Interrupting after reload")
+        Args:
+            timer (int, optional): The time to wait before reloading. Defaults to 0.
 
-            signal.signal(signal.SIGALRM, handler)
-            signal.alarm(10)
+        Raises:
+            RebootSignal: When the device is still unreachable after the timeout period.
```
jmcgill298, line 390:
```suggestion
            RebootTimeoutError: When the device is still unreachable after the timeout period.
```

jmcgill298 (approved):

jmcgill298:
Thanks @mundruid I rebased b/c I saw commits that shouldn't be there, then I realized that the issue was the PR was into the main branch instead of develop. I fixed that, and I updated an incorrect docstring (it wasn't something you changed, just noticed it in the review). I will merge once tests pass

PR #192, opened 2021-01-08
Corrected 'boot_options' regex parsing

qduk:
Line 244 in 'ios_device.py' only captured data from the first capture group. I updated that to be able to get data from either capture group.

### DIFF
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -242,7 +242,7 @@ class IOSDevice(BaseDevice):
         match = re.search(boot_path_regex, show_boot_out, re.MULTILINE)
         if match:
             boot_path_tuple = match.groups()
-            # The regex match will return a tuple of the value and None
+            # The regex match will return two values: the boot value and None
             # The return order will depend on which side of the `or` is matched in the regex
             boot_path = [value for value in boot_path_tuple if value is not None][0]
             file_system = self._get_file_system()
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -242,6 +242,8 @@ class IOSDevice(BaseDevice):
         match = re.search(boot_path_regex, show_boot_out, re.MULTILINE)
         if match:
             boot_path_tuple = match.groups()
+            # The regex match will return a tuple of the value and None
+            # The return order will depend on which side of the `or` is matched in the regex
             boot_path = [value for value in boot_path_tuple if value is not None][0]
             file_system = self._get_file_system()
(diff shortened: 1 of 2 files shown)


pyntc/devices/ios_device.py
```diff
@@ -241,7 +241,8 @@ def boot_options(self):
 
         match = re.search(boot_path_regex, show_boot_out, re.MULTILINE)
         if match:
-            boot_path = match.group(1)
+            boot_path_tuple = match.groups()
+            boot_path = [value for value in boot_path_tuple if value is not None][0]
```
mundruid, line 245: I like the comprehension! Is this always going to be the first not None? Maybe you can exit earlier if it is the first element and not form a whole list?

jvanderaa: Add in a comment to describe what you are doing there?

test/unit/test_devices/test_ios_device.py
```diff
@@ -21,6 +21,35 @@
...
+SHOW_BOOT_VARIABLE = (
+    "Current Boot Variables:\n"
+    "BOOT variable = flash:/cat3k_caa-universalk9.16.11.03a.SPA.bin;\n\n"
+    "Boot Variables on next reload:\n"
+    f"BOOT variable = flash:/{BOOT_IMAGE};\n"
+    "Manual Boot = no\n"
+    "Enable Break = no\n"
+    "Boot Mode = DEVICE\n"
+    "iPXE Timeout = 0"
+)
```
jvanderaa, line 33: Should this be in a file rather than individual lines inside of the test file?

pyntc/devices/ios_device.py
jmcgill298, line 245:
```suggestion
            # The regex match will return a tuple of the value and None
            # The return order will depend on which side of the `or` is matched in the regex
            boot_path = [value for value in boot_path_tuple if value is not None][0]
```

jmcgill298: @mundruid the list comp will always return a list of one value

pyntc/devices/ios_device.py
```diff
@@ -241,7 +241,10 @@ def boot_options(self):
 
         match = re.search(boot_path_regex, show_boot_out, re.MULTILINE)
         if match:
-            boot_path = match.group(1)
+            boot_path_tuple = match.groups()
+            # The regex match will return a tuple of the value and None
```
jmcgill298, line 245:
```suggestion
            # The regex match will return two values: the boot value and None
```

PR #193, opened 2021-01-11
Add Install Mode Option to Install OS

jvanderaa:
Adds an Install Mode flag option to pass through to PyNTC


jmcgill298:
@jvanderaa I think most of these changes are already in the code base. If there are changes, then you will need to rebase against devel.

PR #195, opened 2021-01-14
Merge config() and config_list() for eos.

jeffkala:
Add deprecation for config_list() and update config() to handle both str, list of commands.

Did spin up a vEOS and tested.

### DIFF
--- a/test/unit/test_devices/test_eos_device.py
+++ b/test/unit/test_devices/test_eos_device.py
@@ -27,21 +27,27 @@ class TestEOSDevice(unittest.TestCase):
         # Reset the mock so we don't have transient test effects
         self.device.native.reset_mock()
 
-    def test_config_as_str(self):
+    def test_config_pass_string(self):
         command = "interface Eth1"
         result = self.device.config(command)
 
         self.assertIsNone(result)
         self.device.native.config.assert_called_with(command)
 
-    def test_config_as_list(self):
+    def test_config_pass_list(self):
         commands = ["interface Eth1", "no shutdown"]
         result = self.device.config(commands)
 
         self.assertIsNone(result)
         self.device.native.config.assert_called_with(commands)
 
-    def test_bad_config_as_str(self):
+    @mock.patch.object(EOSDevice, "config")
+    def test_config_list(self, mock_config):
+        commands = ["interface Eth1", "no shutdown"]
+        self.device.config_list(commands)
+        self.device.config.assert_called_with(commands)
+
+    def test_bad_config_pass_string(self):
         command = "asdf poknw"
         response = "Error [1002]: asdf_poknw failed [None]"
 
@@ -50,7 +56,7 @@ class TestEOSDevice(unittest.TestCase):
         assert err.value.command == command
         assert err.value.cli_error_msg == response
 
-    def test_bad_config_as_list(self):
+    def test_bad_config_pass_list(self):
         commands = ["interface Eth1", "apons"]
         response = [
             "Valid",
--- a/test/unit/test_devices/test_eos_device.py
+++ b/test/unit/test_devices/test_eos_device.py
@@ -41,9 +41,6 @@ class TestEOSDevice(unittest.TestCase):
         self.assertIsNone(result)
         self.device.native.config.assert_called_with(commands)
 
-
-    def test_config_list(self):
-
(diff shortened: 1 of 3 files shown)


test/unit/test_devices/test_eos_device.py
```diff
@@ -31,17 +31,17 @@ def test_config(self):
...
 
         self.assertIsNone(result)
-        self.device.native.config.assert_called_with([command])
+        self.device.native.config.assert_called_with(command)
 
     def test_bad_config(self):
         command = "asdf poknw"
 
-        with self.assertRaisesRegex(CommandError, command):
+        with self.assertRaisesRegex(CommandError, command[0]):
```
jmcgill298, line 39: The file needs to import pytest, and this should use pytest to validate exceptions. There should be 2 test cases for the cases CommandError and CommandListError are raised.

An example of using pytest and using the exceptions attributes can be found here: https://github.com/networktocode/pyntc/blob/4e56dfcede757a393e0d1caaa44c46539d949362/test/unit/test_devices/test_base_netmiko.py#L55-L59

test/unit/test_devices/test_eos_device.py
```diff
@@ -50,7 +50,7 @@ def test_bad_config_list(self):
         commands = ["interface Eth1", "apons"]
 
         with self.assertRaisesRegex(CommandListError, commands[1]):
```
jmcgill298, line 52: No need for this test case since the config_list method now just calls the config method.

test/unit/test_devices/test_eos_device.py
```diff
@@ -31,17 +31,17 @@ def test_config(self):
         result = self.device.config(command)
 
         self.assertIsNone(result)
-        self.device.native.config.assert_called_with([command])
```
jmcgill298, line 34: This test case should be duplicated to test passing both a single command as a string and passing a list of commands

test/unit/test_devices/test_eos_device.py
```diff
@@ -31,17 +31,17 @@ def test_config(self):
...
     def test_bad_config(self):
         command = "asdf poknw"
 
-        with self.assertRaisesRegex(CommandError, command):
+        with self.assertRaisesRegex(CommandError, command[0]):
             self.device.config(command)
 
     def test_config_list(self):
         commands = ["interface Eth1", "no shutdown"]
-        result = self.device.config_list(commands)
```
jmcgill298, line 44: This test case should mock self.device.config method and just assert that using config_list method calls the config method and passing the same list of commands to config that it received. An example can be found here: https://github.com/networktocode/pyntc/blob/4e56dfcede757a393e0d1caaa44c46539d949362/test/unit/test_devices/test_base_netmiko.py#L175-L181

This example will not be exactly the same since this is still using older style test cases, and the mocking is done differently there. The logic should still be the same, just need to swap out the mock object and use a method instead of a function

The reason is that the current test case is testing the functionality of what is now in the config method, which is not the goal of unit tests.

jmcgill298 (changes requested):
Thanks for the PR @jeffkala, there are just a few comments on the test cases

jeffkala:
@jmcgill298, thanks for the review.  Will implement these soon.

test/unit/test_devices/test_eos_device.py
jeffkala: @jmcgill298 when you say mock self.device.config would that be a fixture in this case?  I think I'm getting confused that this class in the test file is using unittest and I'm trying to make the assertions with pytest

jmcgill298: This would not be a fixture since this was built using unittest, and has not been converted to our newer testing strategies. You can just use a decorator like

```
@mock.patch.object(EOSDevice, "config")
def test_config_list(self, mock_config):
    commands = ["interface Eth1", "no shutdown"]
    self.device.config_list(commands)
    self.device.config.assert_called_with(commands)
```

jeffkala:
Tested in a docker container on py38 and astroid installed fine.  It failed in the Travis check for 38.  @jmcgill298 anything else I can check?

jmcgill298 (approved):

jmcgill298:
Thanks @jeffkala I kicked the tests off again for py38 and it passed

PR #196, opened 2021-01-19
Merge config() and config_list() for Juniper.

jeffkala:
Deprecation warning for config_list() and repoint it to config() which now turns str into list, and runs commands through for loop to load them into config.

### DIFF
--- a/pyntc/devices/jnpr_device.py
+++ b/pyntc/devices/jnpr_device.py
@@ -125,6 +125,7 @@ class JunosDevice(BaseDevice):
 
          Raises:
              ConfigLoadError: Issue with loading the command.
+             CommandError: Issue with the command provided, if its a single command, passed in as a string.
              CommandListError: Issue with a command in the list provided.
         """
         if isinstance(commands, str):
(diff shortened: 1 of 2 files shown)


pyntc/devices/jnpr_device.py
```diff
@@ -116,22 +117,36 @@ def close(self):
...
+        """
+        if isinstance(commands, str):
+            commands = [commands]
         try:
             for command in commands:
                 self.cu.load(command, format=format)
-
             self.cu.commit()
         except ConfigLoadError as e:
             raise CommandListError(commands, command, e.message)
```
jmcgill298, line 137: This should check to see if `commands` was passed as a list or a string, and then raise either `CommandError` or `CommandListError` as appropriate.

jmcgill298 (changes requested):
Thanks again for the PR @jeffkala There is one comment on updates, and would also include the same updates to test cases as #195

jeffkala:
@jmcgill298, thanks for the review. Will implement these soon.

jmcgill298 (changes requested):

jmcgill298 (approved):

jmcgill298:
Thanks @jeffkala I made one change to how the config_as_list test case was validating the calls.

Issue #176, opened 2020-12-05
Merge config_list into config method


jmcgill298:
The config_list method should be merged into the config method to simplify the interface

jeffkala:
@jmcgill298, both options should still be allowed correct?  We just want to dynamically determine the data that was sent?  `.config("hostname test")` or `.config(["hostname test"])`.

Was thinking I'd give this one a shot.

jmcgill298:
correct, this is will be handled for all netmiko devices in the `netmiko_base` branch. So this would need to be done in F5, EOS, and Juniper drivers (NXOS will migrate to use netmiko once we have the netmiko base class). You can see the implementation for netmiko here `https://github.com/networktocode/pyntc/blob/4e56dfcede757a393e0d1caaa44c46539d949362/pyntc/devices/base_netmiko.py`

Feel free to book some time with me if you would like to go over it before you get started.

jeffkala:
@jmcgill298 since F5, EOS, JUNOS are merged for the config methods, should we close this issue?  Or should I use this same issue for the show merges?

jmcgill298:
We will close this one, thanks @jeffkala

PR #197, opened 2021-01-25
remove *_list methods completely since they were never implemented

jeffkala:
F5 *_list methods were both Not Implemented so just removed them to move forward on the merge/deprecation of the _list methods.

### DIFF
--- a/pyntc/devices/f5_device.py
+++ b/pyntc/devices/f5_device.py
@@ -375,9 +375,6 @@ class F5Device(BaseDevice):
     def config(self, command):
         raise NotImplementedError
 
-    def config_list(self, commands):
-        raise NotImplementedError
-
     @property
     def uptime(self):
         if self._uptime is None:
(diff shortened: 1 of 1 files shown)


mundruid (commented):
With the same logic why not remove the `config`, `show`, and `startup_config` since they are not implemented? All I am trying to figure out is if we had these methods as place holders for future implementation. Thanks!

jeffkala:
> With the same logic why not remove the `config`, `show`, and `startup_config` since they are not implemented? All I am trying to figure out is if we had these methods as place holders for future implementation. Thanks!

@mundruid , its a fair question.  I should have been more clear, I'm working on #176 which is merging config() and config_list() methods, I've done just that on a few other vendors, but since F5 didn't implement the *_list() methods @jmcgill298  recommended just removing them in this case.

mundruid:
> > With the same logic why not remove the `config`, `show`, and `startup_config` since they are not implemented? All I am trying to figure out is if we had these methods as place holders for future implementation. Thanks!
> 
> @mundruid , its a fair question. I should have been more clear, I'm working on #176 which is merging config() and config_list() methods, I've done just that on a few other vendors, but since F5 didn't implement the *_list() methods @jmcgill298 recommended just removing them in this case.

@jeffkala Makes sense! Thank you for the prompt response.

mundruid (approved):

PR #198, opened 2021-01-26
retrying ap image swapp all command

balmasea:


jmcgill298:
Thanks @balmasea, the `test_transfer_image_to_ap_*` test cases in `test/unit/test_devices/test_aireos_device.py` should be updated to check that the mock_ap_image_matches_expected was called 3 times and a new test case should be created that fails the condition on the first iteration, and then passes on the second iteration.

balmasea:
> Thanks @balmasea, the `test_transfer_image_to_ap_*` test cases in `test/unit/test_devices/test_aireos_device.py` should be updated to check that the mock_ap_image_matches_expected was called 3 times and a new test case should be created that fails the condition on the first iteration, and then passes on the second iteration.

New commits should do it. Thanks.

test/unit/test_devices/test_aireos_device.py
```diff
@@ -1592,16 +1592,64 @@ def test_transfer_image_to_ap_transfer_secondary_fail(
...
     mock_boot_options.assert_has_calls([mock.call(), mock.call()])
     assert fte.value.message == f"Unable to set all APs to use {aireos_boot_image}"
 
 
+@mock.patch.object(AIREOSDevice, "config")
+@mock.patch.object(AIREOSDevice, "_wait_for_ap_image_download")
+@mock.patch.object(AIREOSDevice, "boot_options", new_callable=mock.PropertyMock)
+@mock.patch.object(AIREOSDevice, "_ap_images_match_expected")
+@mock.patch.object(AIREOSDevice, "ap_boot_options", new_callable=mock.PropertyMock)
+def test_transfer_image_to_ap_transfer_fail_at_first_try(
```
jmcgill298, line 1610:
```suggestion
def test_transfer_image_to_ap_transfer_fail_swap_at_first_try(
```

test/unit/test_devices/test_aireos_device.py
```diff
@@ -1592,16 +1592,64 @@ def test_transfer_image_to_ap_transfer_secondary_fail(
...
+    mock_ap_image_matches_expected,
+    mock_boot_options,
+    mock_wait,
+    mock_config,
+    aireos_device,
+    aireos_mock_path,
+    aireos_boot_image,
+):
+    mock_boot_options.return_value = {"primary": None, "backup": aireos_boot_image}
+    mock_ap_image_matches_expected.side_effect = [False, False, False, False, True, True, True, True]
```
jmcgill298, line 1621:
```suggestion
    mock_ap_image_matches_expected.side_effect = [False, False, True, True, True, False, True]
```

test/unit/test_devices/test_aireos_device.py
```diff
@@ -1592,16 +1592,64 @@ def test_transfer_image_to_ap_transfer_secondary_fail(
...
+    assert aireos_device.transfer_image_to_ap(aireos_boot_image) is True
+    assert len(mock_ap_image_matches_expected.mock_calls) == 5
+    mock_config.assert_has_calls([mock.call("ap image predownload backup all")])
+    mock_wait.assert_called()
```
jmcgill298, line 1625:
```suggestion
    assert len(mock_ap_image_matches_expected.mock_calls) == 7
    mock_config.assert_has_calls([mock.call("ap image swap all")] * 2)
```

jmcgill298: Need to adjust the values for testing swap fails on first try

test/unit/test_devices/test_aireos_device.py
jmcgill298: Need to swap return values to cause swap to fail on first iteration

test/unit/test_devices/test_aireos_device.py
```diff
@@ -1592,16 +1592,64 @@ def test_transfer_image_to_ap_transfer_secondary_fail(
...
+    mock_boot_options.return_value = {"primary": None, "backup": aireos_boot_image}
+    mock_ap_image_matches_expected.side_effect = [False, False, False, False, False]
+    with pytest.raises(aireos_module.FileTransferError) as fte:
+        aireos_device.transfer_image_to_ap(aireos_boot_image)
+    assert len(mock_ap_image_matches_expected.mock_calls) == 5
+    mock_wait.assert_called()
+    mock_boot_options.assert_has_calls([mock.call(), mock.call()])
+    assert fte.value.message == f"Unable to set all APs to use {aireos_boot_image}"
+
+
```
jmcgill298, line 1652:
```suggestion
```

This test case is accounted for in other fail test cases

jmcgill298 (commented):
@balmasea I made some updates to the test cases

PR #199, opened 2021-01-26
merge show() and show_list() for eos device

jeffkala:
Merging show and show_list methods for issue 185.  Deprecate message for show_list method.

### DIFF
--- a/pyntc/devices/eos_device.py
+++ b/pyntc/devices/eos_device.py
@@ -395,7 +395,9 @@ class EOSDevice(BaseDevice):
             return response_list
         except EOSCommandError as e:
             if command_list.get("type") == "string":
+                print(f"This is the full exception:\n {e}")
                 raise CommandError(e.commands, e.message)
+            print(f"This is the full exception:\n {e}")
             raise CommandListError(commands, e.commands[len(e.commands) - 1], e.message)
 
     def show_list(self, commands):
--- a/pyntc/devices/eos_device.py
+++ b/pyntc/devices/eos_device.py
@@ -376,25 +376,22 @@ class EOSDevice(BaseDevice):
             CommandError: Issue with the command provided.
             CommandListError: Issue with a command in the list provided.
         """
-        if raw_text:
-            encoding = "text"
-        else:
+        if not raw_text:
             encoding = "json"
-
-        if isinstance(commands, str):
-            command_list = dict(type="string", commands=[commands])
         else:
-            command_list = dict(type="list", commands=commands)
+            encoding = "text"
 
+        original_commands_is_str = isinstance(commands, str)
+        if original_commands_is_str :
+            commands=[commands]
         try:
-            response_list = self._parse_response(
-                self.native.enable(command_list.get("commands"), encoding=encoding), raw_text=raw_text
-            )
-            if command_list.get("type") == "string":
+            response = self.native.enable(commands, encoding=encoding)
+            response_list = self._parse_response(response, raw_text=raw_text )
+            if original_commands_is_str:
                 return response_list[0]
             return response_list
         except EOSCommandError as e:
-            if command_list.get("type") == "string":
+            if original_commands_is_str:
                 raise CommandError(e.commands, e.message)
             raise CommandListError(commands, e.commands[len(e.commands) - 1], e.message)
 
--- a/pyntc/devices/eos_device.py
+++ b/pyntc/devices/eos_device.py
@@ -395,9 +395,7 @@ class EOSDevice(BaseDevice):
             return response_list
         except EOSCommandError as e:
             if command_list.get("type") == "string":
-                print(f"This is the full exception:\n {e}")
                 raise CommandError(e.commands, e.message)
-            print(f"This is the full exception:\n {e}")
             raise CommandListError(commands, e.commands[len(e.commands) - 1], e.message)
 
     def show_list(self, commands):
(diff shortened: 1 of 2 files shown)


pyntc/devices/eos_device.py
```diff
@@ -364,25 +365,49 @@ def set_boot_options(self, image_name, **vendor_specifics):
...
+        if isinstance(commands, str):
+            command_list = dict(type="string", commands=[commands])
+        else:
+            command_list = dict(type="list", commands=commands)
+
         try:
-            return self._parse_response(self.native.enable(commands, encoding=encoding), raw_text=raw_text)
+            response_list = self._parse_response(
+                self.native.enable(command_list.get("commands"), encoding=encoding), raw_text=raw_text
+            )
```
jmcgill298, line 392:
```suggestion
            response = self.native.enable(commands, encoding=encoding)
            response_list = self._parse_response(response, raw_text=raw_text )
```
I formatted this per my suggestion above, but regardless of what is accepted there, this should be split into two separate steps

pyntc/devices/eos_device.py
```diff
@@ -364,25 +365,49 @@ def set_boot_options(self, image_name, **vendor_specifics):
...
+            if command_list.get("type") == "string":
```
jmcgill298, line 393:
```suggestion
            if original_commands_is_str:
```
Would need to be updated if above suggestion is agreed upon

pyntc/devices/eos_device.py
```diff
@@ -364,25 +365,49 @@ def set_boot_options(self, image_name, **vendor_specifics):
...
+                return response_list[0]
+            return response_list
         except EOSCommandError as e:
+            if command_list.get("type") == "string":
```
jmcgill298, line 397:
```suggestion
            if original_commands_is_str:
```
Would need to be updated if above suggestion is agreed upon

test/unit/test_devices/test_eos_device.py
```diff
@@ -62,10 +63,36 @@ def test_show(self):
```
jmcgill298, line 63: These 4 test cases should mock `_parse_response` method and assert that it was called/not called as appropriate

jmcgill298 (changes requested):
Thanks @jeffkala, I have a few comments on this one, mostly regarding updates what existed before. The one thing I did mention about this implementation is regarding the logic for handling list vs string implementations. I put in suggestions for what reads easier to me.

test/unit/test_devices/test_eos_device.py
jeffkala: I updated them, but only 3 of the 5, the 2 test that test the raise exception I couldn't understand why those would need the mock since it would fail before then.

test/unit/test_devices/test_eos_device.py
```diff
@@ -52,45 +53,85 @@ def test_bad_config_list(self):
...
+                        {"hwAddress": "2cc2.60ff.0011", "interface": "Management1", "age": 0, "address": "10.0.0.2"}
+                    ],
+                    "notLearnedEntries": 0,
+                    "totalEntries": 1,
+                    "dynamicEntries": 1,
+                    "staticEntries": 0,
+                },
+                "encoding": "json",
+            }
+        ]
```
jmcgill298, line 73:
```suggestion
        return_value = [
            {
                "command": "show ip arp",
                "result": {
                    "ipV4Neighbors": [
                        {"hwAddress": "2cc2.60ff.0011", "interface": "Management1", "age": 0, "address": "10.0.0.2"}
                    ],
                    "notLearnedEntries": 0,
                    "totalEntries": 1,
                    "dynamicEntries": 1,
                    "staticEntries": 0,
                },
                "encoding": "json",
            }
        ]
        mock_parse.return_value = return_value
```

test/unit/test_devices/test_eos_device.py
```diff
@@ -52,45 +53,85 @@ def test_bad_config_list(self):
...
         result = self.device.show(command)
-
-        self.assertIsInstance(result, dict)
-        self.assertNotIn("command", result)
-        self.assertIn("dynamicEntries", result)
-
+        self.device._parse_response.assert_called_with([result], raw_text=False)
```
jmcgill298, line 75:
```suggestion
        assert result == return_value[0]
        self.device._parse_response.assert_called_with([result], raw_text=False)
```

test/unit/test_devices/test_eos_device.py
```diff
@@ -52,45 +53,85 @@ def test_bad_config_list(self):
...
+                "result": {"hostname": "eos-spine1", "fqdn": "eos-spine1.ntc.com"},
+                "encoding": "json",
+            },
+            {
+                "command": "show clock",
+                "result": {"output": "Fri Jan 22 23:29:21 2016\nTimezone: UTC\nClock source: local\n"},
+                "encoding": "text",
+            },
+        ]
+        result = self.device.show(commands)
```
jmcgill298, line 93:
```suggestion
        return_value = [
            {
                "command": "show hostname",
                "result": {"hostname": "eos-spine1", "fqdn": "eos-spine1.ntc.com"},
                "encoding": "json",
            },
            {
                "command": "show clock",
                "result": {"output": "Fri Jan 22 23:29:21 2016\nTimezone: UTC\nClock source: local\n"},
                "encoding": "text",
            },
        ]
        mock_parse.return_value = return_value
        result = self.device.show(commands)
        assert result == return_value
```

jmcgill298:
Looks good @jeffkala After reviewing the updates, I realized that I led you astray on not validating the result of the method. I forgot that the code returned the first result or the whole result based on passing a string or a list, which we would want to test. I added that back and will merge once Travis passes

jmcgill298 (approved):

PR #206, opened 2021-02-23
Added docstrings to eos_device.py

qduk:


pyntc/devices/eos_device.py
```diff
@@ -193,6 +220,11 @@ def enable(self):
 
     @property
     def uptime(self):
```
mundruid, line 222: This property assigns only and does not explicitly return. I think it would be better to omit the `Returns` statement in the docstring.

pyntc/devices/eos_device.py
```diff
@@ -201,13 +233,23 @@ def uptime(self):
 
     @property
     def uptime_string(self):
+        """Create a string of the uptime of the device.
+
+        Returns:
```
mundruid, line 238: This property assigns only and does not explicitly return. I think it would be better to omit the `Returns` statement in the docstring.

pyntc/devices/eos_device.py
```diff
@@ -201,13 +233,23 @@ def uptime(self):
...
         if self._uptime_string is None:
             self._uptime_string = self._uptime_to_string(self.uptime)
 
         return self._uptime_string
 
     @property
     def hostname(self):
+        """Get hostname from device.
+
+        Returns:
```
mundruid, line 250: Same here: This property assigns only and does not explicitly return. I think it would be better to omit the `Returns` statement in the docstring.

mundruid: This applies to all the following properties that do not explicitly return.

pyntc/devices/eos_device.py
```diff
@@ -298,6 +370,16 @@ def file_copy(self, src, dest=None, file_system=None):
...
     def file_copy_remote_exists(self, src, dest=None, file_system=None):
+        """Copy file to remote device if it exists.
+
+        Args:
+            src (string): source file
+            dest (string, optional): Destintion file. Defaults to None.
+            file_system (string, optional): Describes device file system. Defaults to None.
+
+        Returns:
+            [type]: [description]
```
mundruid, line 381: Missed that one: `bool: True if remote file exists.`

mundruid (changes requested):
I noticed that you are using the `Returns` statement in properties that do not explicitly return. This is not a big error, but it would be better not to have the `Returns` statement when a function does not return something.

pyntc/devices/eos_device.py
```diff
@@ -136,19 +146,35 @@ def _wait_for_device_reboot(self, timeout=3600):
         raise RebootTimeoutError(hostname=self.hostname, wait_time=timeout)
 
     def backup_running_config(self, filename):
+        """Create backup file of running configurtation.
```
jmcgill298, line 149:
```suggestion
        """
        Create backup file of running configuration.
```

pyntc/devices/eos_device.py
```diff
@@ -201,13 +233,23 @@ def uptime(self):
 
     @property
     def uptime_string(self):
+        """Create a string of the uptime of the device.
```
jmcgill298, line 236:
```suggestion
        """
        Get uptime of the device in the format of dd::hh::mm.
```

pyntc/devices/eos_device.py
```diff
@@ -193,6 +220,11 @@ def enable(self):
 
     @property
     def uptime(self):
+        """Get uptime from device.
```
jmcgill298, line 223:
```suggestion
        """
        Get uptime of the device in seconds.
```

jmcgill298 (changes requested):

jmcgill298 (approved):

PR #214, opened 2021-03-27
Fixes proper boot system setting on IOS devices

### LINKED ISSUE #213: Unable to properly set boot image on old Cisco IOS devices
Had an issue with Cisco ISR 2901 during the upgrade.  The proper syntax on the device is `boot system <FS> <IMAGE_NAME>`. 
The code first tries to implement `boot system <FS>:/<IMAGE_NAME>` and, if that fails, falls back to `boot system <FS> <IMAGE_NAME>` command. Strangely enough, the device does not complain about the first command and subsequent confirmation of the boot setting fails.

To replicate:
```
>>> from pyntc.devices.ios_device import IOSDevice
>>> d = IOSDevice(host=ISR2901, username='admin', password='password')
>>> d.install_os('c2900-universalk9-mz.SPA.154-3.M10.bin')
  File "<stdin>", line 1, in <module>
  File "/home/user/.local/lib/python3.8/site-packages/pyntc/devices/ios_device.py", line 860, in set_boot_options
  raise CommandError(
                command=command,
                message="Setting boot command did not yield expected results, found {0}".format(new_boot_options),
            )
pyntc.errors.OSInstallError: CommandError:
Command boot system flash0:/c2900-universalk9-mz.SPA.154-3.M10.bin was not successful: Setting boot command did not yield expected results, found boot-end-marker
>>> print(d.show('show run | in boot))
'ISR2901#show run  |  in boot
boot-start-marker
boot system flash0:/c2900-universalk9-mz.SPA.154-3.M10.bin
boot-end-marker
no ip bootp server
'

### PR DESCRIPTION
**Pull request by armartirosyan**
Used regex to determine the current syntax and based on it issued proper config command.
Fixes #213

### DIFF
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -249,7 +249,7 @@ class IOSDevice(BaseDevice):
             except CommandError:
                 # Default to running config value
                 show_boot_out = self.show("show run | inc boot")
-                boot_path_regex = r"boot\s+system\s+(?:\S+\s+|)(\S+)\s*$"
+                boot_path_regex = r"boot\ssystem\s\S+(?::/+|\s)(\S+.bin)"
 
         match = re.search(boot_path_regex, show_boot_out, re.MULTILINE)
         if match:
@@ -616,7 +616,12 @@ class IOSDevice(BaseDevice):
                         self.show(command, delay_factor=install_mode_delay_factor)
                     except IOError:
                         pass
-
+                    except CommandError:
+                        command = (
+                            f"request platform software package install switch all file {self._get_file_system()}{image_name} auto-copy"
+                        )
+                        self.show(command, delay_factor=install_mode_delay_factor)
+                        self.reboot()
             else:
                 self.set_boot_options(image_name, **vendor_specifics)
                 self.reboot()
@@ -627,7 +632,7 @@ class IOSDevice(BaseDevice):
             # Set FastCLI back to originally set when using install mode
             if install_mode:
                 self.fast_cli = current_fast_cli
-
+                image_name = INSTALL_MODE_FILE_NAME
             # Verify the OS level
             if not self._image_booted(image_name):
                 raise OSInstallError(hostname=self.hostname, desired_boot=image_name)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by jmcgill298 — CHANGES_REQUESTED** [review]
Thanks @armartirosyan 

We do need an updated test case so we can avoid a later regression back to this issue.

File: pyntc/devices/ios_device.py [hunk 1]
```diff
@@ -846,13 +846,22 @@ def set_boot_options(self, image_name, **vendor_specifics):
...
-        except CommandListError:  # TODO: Update to CommandError when deprecating config_list
+        elif re.search(r'boot\ssystem\s\S+\s\S+', boot_sys_syntax): # Will match boot system flash cat3k_caa-universalk9.16.12.01.SPA.bin
             file_system = file_system.replace(":", "")
             command = "boot system {0} {1}".format(file_system, image_name)
             self.config(["no boot system", command])
+        else:
+            raise CommandError(
+                command=command,
+                message="Unable to determine the proper boot system syntax. Current configuration is {0}".format(boot_sys_syntax)
+            )
```
jvanderaa, line 864: The logic is making sense, but this will miss on one scenario, the Install Mode devices. A Cisco 9300 does not have a boot statement in the configuration. 

We may be able to do a check if the device is using install mode. I'm just not certain the module has the data yet.

jvanderaa (changes requested):
We should work on the changes yet. What you are working towards makes sense, but the run command won't catch everything.

pyntc/devices/ios_device.py
armartirosyan: @jvanderaa - I updated the logic to work with both bundle and install modes. 
Thanks for the feedback.

balmasea:
Hi @armartirosyan! Is this PR dead?

jeffkala:
@armartirosyan big delay on reaching back out, any interest in taking this over the finish line?

armartirosyan:
Hey @jeffkala, I totally forgot about this PR. 😔Yeah absolutely. I think the only thing left is to update the tests. I will do my best to address it this weekend.

armartirosyan:
Hey @jeffkala,
I fixed the unittest and some other things and tired to commit, but looks like I don't have permissions to  commit directly to my branch
```bash
% git push origin armen_issue213_ios_code_upgrade_syntax
ERROR: Permission to networktocode/pyntc.git denied to armartirosyan.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
I think I need to close this PR, fork the repo, make my changes, then create a new PR...

PR #215, opened 2021-03-30
Added pydocstyle to tox

qduk:
There were also a few things that needed cleaning for pydocstyle to pass.


qduk:
@jmcgill298 Can you please double check this function `def get_structured_data(template_name, rawtxt):` and the functions in `converters.py` when your review? I had to add docstrings there but want to make sure I captured pertinent information.

jmcgill298:
Thanks @qduk

PR #219, opened 2021-08-14
Feature/rollback depending on pwd

jmcgill298:


pyntc/devices/ios_device.py
```diff
@@ -922,7 +922,7 @@ def rollback(self, rollback_to):
             RollbackError: Error if unable to rollback to configuration.
         """
         try:
-            self.show("configure replace flash:%s force" % rollback_to)
+            self.show("configure replace %s%s force" % (self._get_file_system(), rollback_to))
```
mundruid, line 925: ` self.show("configure replace %s %s force" % (self._get_file_system(), rollback_to))`

Should we add a space between the two strings or do you want to present these as one?

mundruid (commented):
LGTM I just had a cosmetic comment.

pyntc/devices/ios_device.py
jmcgill298: It is a single string, like `flash:ios-15.2advipservices.bin`

PR #233, opened 2022-03-10
Accept a higher native timeout in install os for nexus.

### LINKED ISSUE #232: N5Ks can be slow the native timeout should be configurable.
https://github.com/networktocode/pyntc/blob/b46b94a202c1dc5bf30c97dcaa69af8fe77bde60/pyntc/devices/nxos_device.py#L365

Constantly hitting the `requests` module `read timeout` which is 300 on N5K installs.  These installs go through the validation of both the system and kickstart image.  Regularity takes longer than 10 minutes to just finish the "verification" steps.  This should be configurable to avoid timeout issues.

jeffkala:
fixes #232

### DIFF
--- a/pyntc/devices/nxos_device.py
+++ b/pyntc/devices/nxos_device.py
@@ -4,19 +4,19 @@ import re
 import time
 import warnings
 
-from .base_device import BaseDevice, RollbackError, RebootTimerError, fix_docs
 from pyntc.errors import (
     CommandError,
-    OSInstallError,
     CommandListError,
     FileTransferError,
-    RebootTimeoutError,
     NTCFileNotFoundError,
+    OSInstallError,
+    RebootTimeoutError,
 )
-
 from pynxos.device import Device as NXOSNative
-from pynxos.features.file_copy import FileTransferError as NXOSFileTransferError
 from pynxos.errors import CLIError
+from pynxos.features.file_copy import FileTransferError as NXOSFileTransferError
+
+from .base_device import BaseDevice, RebootTimerError, RollbackError, fix_docs
 
 
 @fix_docs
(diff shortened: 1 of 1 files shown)


pyntc/devices/nxos_device.py
```diff
@@ -362,7 +362,8 @@ def set_boot_options(self, image_name, kickstart=None, **vendor_specifics):
             kickstart = file_system + kickstart
 
         image_name = file_system + image_name
-        self.native.timeout = 300
+        if self.native.timeout < 300:
```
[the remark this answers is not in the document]
jeffkala: Added a comment.  Essentially just to support a user defined timeout if its higher than 300.  I'd say if a user is defining a higher timeout it is for a reason.  Slow switch, or in my case longer upgrade cycle for n5k hardware.

mundruid (approved):
Thank you!

jeffkala:
Let me know if anything else is needed.

pyntc/devices/nxos_device.py
```diff
@@ -362,7 +362,9 @@ def set_boot_options(self, image_name, kickstart=None, **vendor_specifics):
             kickstart = file_system + kickstart
 
         image_name = file_system + image_name
-        self.native.timeout = 300
+        # Allow for user defined timeout to take precedence if its over 300 seconds, otherwise change to 300.
+        if self.native.timeout < 300:
```
jmcgill298, line 366: what if self.native.timeout isn't an int?

jmcgill298 (changes requested):

pyntc/devices/nxos_device.py
jeffkala: This has been updated @jmcgill298

mundruid (approved):

PR #262, opened 2022-10-31
Fixes proper boot system setting on IOS devices

armartirosyan:
Used regex to determine the current syntax and based on it issued proper config command.
Fixes https://github.com/networktocode/pyntc/issues/213

Original PR: https://github.com/networktocode/pyntc/pull/214


balmasea:
I think this PR solves issue #263 too. Nice!

jeffkala (approved):

Issue #266, opened 2022-11-25
Upgrade in install mode does not wait for reboot


balmasea:
### Environment
* Python version: 3.6.8
* pyntc version:  0.20.2

### Expected Behavior
After running an upgrade in install mode, I would expect that library would wait for the device to be reloaded.
```
Cisco IOS XE Software, Version 17.06.03
Cisco IOS Software [Bengaluru], Catalyst L3 Switch Software (CAT9K_IOSXE), Version 17.6.3, RELEASE SOFTWARE (fc4)
Technical Support: http://www.cisco.com/techsupport
Copyright (c) 1986-2022 by Cisco Systems, Inc.
Compiled Wed 30-Mar-22 23:09 by mcpre


Cisco IOS-XE software, Copyright (c) 2005-2022 by cisco Systems, Inc.
All rights reserved.  Certain components of Cisco IOS-XE software are
licensed under the GNU General Public License ("GPL") Version 2.0.  The
software code licensed under GPL Version 2.0 is free software that comes
with ABSOLUTELY NO WARRANTY.  You can redistribute and/or modify such
GPL code under the terms of GPL Version 2.0.  For more details, see the
documentation or "License Notice" file accompanying the IOS-XE software,
or the applicable URL provided on the flyer accompanying the IOS-XE
software.


ROM: IOS-XE ROMMON
BOOTLDR: System Bootstrap, Version 16.12.2r, RELEASE SOFTWARE (P)

hostname uptime is 2 days, 14 hours, 55 minutes
Uptime for this control processor is 2 days, 14 hours, 56 minutes
System returned to ROM by Reload Command at 21:00:42 UTC Tue Nov 22 2022
System restarted at 21:03:09 UTC Tue Nov 22 2022
System image file is "flash:/packages.conf"
Last reload reason: Reload Command
```

### Observed Behavior
It immediately runs the method `_wait_for_device_reboot`, capturing the `show version` output and exception `OSInstallError` is raised.
Show version output fragment:
```Cisco IOS XE Software, Version 16.12.04
Cisco IOS Software [Gibraltar], Catalyst L3 Switch Software (CAT9K_IOSXE), Version 16.12.4, RELEASE SOFTWARE (fc5)
Technical Support: http://www.cisco.com/techsupport
Copyright (c) 1986-2020 by Cisco Systems, Inc.
Compiled Thu 09-Jul-20 21:49 by mcpre


[...]

PR #268, opened 2022-12-15
Feature/verifying reload has happened

balmasea:
Checking that indeed reload does not happen fast enough for some devices, I have implemented a slightly rudimentary fix that waits for the reload for ten minutes max.

This should solve https://github.com/networktocode/pyntc/issues/266 or at least, put us in the correct direction.

### DIFF
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -229,6 +229,9 @@ class IOSDevice(BaseDevice):
 
         raise RebootTimeoutError(hostname=self.hostname, wait_time=timeout)
 
+    def _has_reload_happened_recently(self):
+        return re.search("^00:00:0\d:*", self.uptime_string) is not None
+
     def backup_running_config(self, filename):
         """Backup running configuration to filename specified.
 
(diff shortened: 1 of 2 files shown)


jeffkala:
@balmasea thanks for the continued contributions. Will be reviewing this shortly.

pyntc/devices/ios_device.py
```diff
@@ -707,6 +710,11 @@ def install_os(self, image_name, install_mode=False, install_mode_delay_factor=2
                     # Set a higher delay factor and send it in
                     try:
                         self.show(command, delay_factor=install_mode_delay_factor)
+                        MAXIMUM_RETRIES = 10
```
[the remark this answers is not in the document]
balmasea: Agree, I like very much that approach.

pyntc/devices/ios_device.py
```diff
@@ -222,13 +222,16 @@ def _wait_for_device_reboot(self, timeout=3600):
         while time.time() - start < timeout:
             try:
                 self.open()
-                self.show("show version")
```
jeffkala, line 225: you would need `self.show("show version")` still right?

balmasea: If I'm not wrong, it is still being executed when getting the `uptime_string`. I think a need a different way of getting it though.

balmasea: This commit should make it work https://github.com/networktocode/pyntc/pull/268/commits/c31aebacdce62bf011e2c8df5e449cab53265920

jeffkala: Thanks @balmasea , this looks like a good option.

jeffkala (approved):

PR #271, opened 2022-12-22
Add logging for all devices Taking over the finish line

jeffkala:


pyntc/devices/ios_device.py
```diff
@@ -222,11 +236,14 @@ def _wait_for_device_reboot(self, timeout=3600):
         while time.time() - start < timeout:
             try:
                 self.open()
+                self.show("show version")
+                log.debug("Host %s: Device rebooted.", self.host)
```
balmasea, line 240: Sorry to jump here, I was having a look at the PR: bear in mind that this may not be true, as it exists the possibility that the device opens the connection and returns a `show version` before being rebooted.

Issue #183, opened 2020-12-20
Add "facts" properties to ASA and AireOS


jmcgill298:
Both AireOS and ASA device classes need to implement properties that were previously associated with facts:

ASA:
- [ ] vendor
- [ ] uptime
- [ ] uptime_string
- [ ] hostname
- [ ] fqdn
- [ ] interfaces
- [ ] vlans
- [ ] model
- [ ] serial_number
- [ ] os_version

AIREOS:
- [ ] vendor
- [ ] uptime
- [ ] uptime_string
- [ ] hostname
- [ ] fqdn
- [ ] interfaces
- [ ] vlans
- [ ] model
- [ ] serial_number
- [ ] os_version

PR #275, opened 2023-04-05
Remove ABC, deprecate and remove show_list and config_list methods, add fact properties

### LINKED ISSUE #96: Remove ABCs
I do not believe we need to use Abstract Base Classes for this project. We could probably refactor how inheritance works.

jeffkala:
This implements most of #183 , closes #96 , removes all config_list and show_list methods completely (hence a major version bump is needed.


jeffkala:
@jmcgill298 think this should cover some of the long outstanding issues, would love a review if you get a chance.

jmcgill298:
looks good @jeffkala

PR #289, opened 2023-06-01
Feature/boot swicth all

juanjtomasg:
C3650 devices do have a non matching syntax in current pyntc code.

C3650 #sh run | i boot system
boot system switch all flash:cat3k_caa-universalk9.SPA.03.07.01.E.152-3.E1.bin

The issue is that the current regex expression do not allow to have more than one string without spaces. Adding new regex expresion that allows to have the string "swicth all" within the boot system command.


balmasea:
Hi @jeffkala, @pszulczewski and @pke11y. Could you please review this PR? We believe it is interesting for the library.
Thanks.

pszulczewski (approved):
LGTM

PR #300, opened 2023-11-28
Fix default ports.

pszulczewski:
When using `nautobot-nornir` the default value of device port is `None`.

```
In [16]: nornir_obj.inventory.hosts['3750-x'].dict()
Out[16]: 
{'name': '3750-x',
 'connection_options': {'netmiko': {'extras': {'secret': None},
   'hostname': None,
   'port': None,
   'username': None,
   'password': None,
   'platform': 'cisco_ios'},
    ...
```

This causes an error on `int(port)` in the constructor:
```
nautobot_os_upgrades-worker-1    |   File "/usr/local/lib/python3.8/site-packages/pyntc/devices/ios_device.py", line 65, in __init__
nautobot_os_upgrades-worker-1    |     self.port = int(port)
nautobot_os_upgrades-worker-1    | TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

This PR fixes this behavior, by allowing `port=None` to be passed, but still falls back to the default value of 22.
Port default value in the constructor has been changed fro `22` to `None` for uniformity with other drivers.
This change also eliminates the need to set ports in nautobot plugin configuration options.


jeffkala:
can you drop py3.7?

pszulczewski:
> can you drop py3.7?

Yes, just realized that a while ago. It's done.

jeffkala (approved):

PR #307, opened 2023-12-01
Test only the first and last python version in GHA

pszulczewski:


whitej6 (approved):

jeffkala (approved):

pszulczewski:
Will add that in #308

PR #323, opened 2025-02-03
Fix/issue 317

### LINKED ISSUE #317: Install OS for IOS goes to infinite loop when there is not enough free disk to complete an install mode upgrade
### Environment
* Python version:  3.9.18
* pyntc version:  2.0.0

### Expected Behavior
After running a command 
```# install add file flash:cat9k_iosxe.17.12.04.SPA.bin activate commit prompt-level none``` 
and getting this output
`FAILED: There is not enough free disk available to perform this operation on switch 1. At least 1276287 KB of free disk is required`
error should be reported and process should be interrupted.

### Observed Behavior
Instead of early failing, it is waiting for the device reboot until timeout expires.

### Steps to Reproduce
1. Run `IOSDevice::install_os` with install_mode as True in a device with not enough free disk available.
2. Check that commands `show version` is being executed in a loop until timeout expires.

balmasea:
Fixing issue #317

### DIFF
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -718,7 +718,10 @@ class IOSDevice(BaseDevice):
                     )
                     # Set a higher read_timeout and send it in
                     try:
-                        self.show(command, read_timeout=read_timeout)
+                        install_message = self.show(command, read_timeout=read_timeout)
+                        if install_message.startswith("FAILED:"):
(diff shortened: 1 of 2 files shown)


pszulczewski (approved):
Many thanks for adding this fix.
It looks good to me.

jeffkala (approved):

PR #332, opened 2025-03-27
Housekeeping: Fix CI Pipeline - updating GHA to supported versions

### LINKED ISSUE #247: Prep for deprecation of py37 as prep work for eol  Mid2023

### LINKED ISSUE #331: CI Pipeline Failures due to dependency issues for Actions/Cache
### Environment
* Python version: all, 3.11, 3.13
* pyntc version:  all, 2.0.2

### Expected Behavior

CI within Github runs and passes for known good code. 

### Observed Behaviour

CI continually fails

### Steps to Reproduce
1. Create PR
2. CI will fail

### issue 

From - https://github.com/actions/cache
What's New

⚠️ Important changes

The cache backend service has been rewritten from the ground up for improved performance and reliability. [actions/cache](https://github.com/actions/cache) now integrates with the new cache service (v2) APIs.

The new service will gradually roll out as of February 1st, 2025. The legacy service will also be sunset on the same date. Changes in these release are fully backward compatible.

We are deprecating some versions of this action. We recommend upgrading to version v4 or v3 as soon as possible before February 1st, 2025. (Upgrade instructions below).

If you are using pinned SHAs, please use the SHAs of versions v4.2.0 or v3.4.0

If you do not upgrade, all workflow runs using any of the deprecated [actions/cache](https://github.com/actions/cache) will fail.

Upgrading to the recommended versions will not break your workflows.

### LINKED ISSUE #333: EoL Python3.8
### Environment
* pyntc version: 2.0.2

Python 3.8 is End of Life and we should develop against current versions, suggested 3.11
### Proposed Functionality

Update tools, housekeeping to ensure 3.11 is the base standard.

### Use Case

Keeping project current

garymccann:
Fixes #331 
Closes #247
Closes #333 

From 1st Feb - know issues with existing actions that depend on Cache. I have updated all the GHA to latest versions. 

Side Effect - I have went ahead and upgraded code to be 3.11 compatible. This should not be a breaking change for 3.9+, but feedback appreciated on whether this will cause issues for any users.


.github/workflows/ci.yml
```diff
@@ -83,22 +83,22 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.8", "3.11"]
+        python-version: ["3.11", "3.12"]
```
jeffkala: should be the lowest and the highest versions we support.

.github/workflows/ci.yml
```diff
@@ -155,22 +155,22 @@ jobs:
...
```
jeffkala: same as previous should be `["3.9", "3.12"]`

pyproject.toml
```diff
@@ -13,18 +13,18 @@ classifiers = [
...
+    "Programming Language :: Python :: 3.12",
 ]
 include = [
     "LICENSE",
     "README.md",
 ]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "^3.11"
```
jeffkala: Are we support 3.9?  I don't think we want to drop 3.9 and 3.10 yet.

Change to `^3.9`

pyproject.toml
```diff
@@ -69,7 +69,7 @@ file = "README.md"
 
 [tool.black]
 line-length = 120
-target-version = ['py37']
+target-version = ['py311']
```
jeffkala: change to lowest version we support py39

jeffkala (changes requested):

jeffkala:
> Fixes #331 Closes #247 Closes #333
> 
> From 1st Feb - know issues with existing actions that depend on Cache. I have updated all the GHA to latest versions.
> 
> Side Effect - I have went ahead and upgraded code to be 3.11 compatible. This should not be a breaking change for 3.9+, but feedback appreciated on whether this will cause issues for any users.

We should be keeping py3.9 support until its EOL.

pyproject.toml
garymccann: Done

pyproject.toml
garymccann: Done

jeffkala (approved):

PR #337, opened 2025-07-17
fix docs build and docs around standards, towncrier, etc.

jeffkala:


mkdocs.yml
```diff
@@ -88,7 +89,10 @@ plugins:
         python:
           paths: ["."]
           options:
+            heading_level: 1
             show_root_heading: true
+            show_root_members_full_path: true
+            show_source: false
```
itdependsnetworks, line 119: Look below for:

```
plugins:
  - "search"
  - "mkdocs-version-annotations"
```

and change to

https://github.com/nautobot/nautobot/blob/ba2b3028566b1995f4ca868404f8e6aec116d458/mkdocs.yml#L82-L84

jeffkala: I'm really not following what you mean on this.

jeffkala: @itdependsnetworks can you expand. The linked info doesn't seem to match what the reference code is.

itdependsnetworks: I can't comment on the code I want to since it wasn't change, but I am advocating to change the code that says "mkdocs-version-annotations" to the linked code.

jeffkala: Did an overall pass of getting the mkdocs plugins/extensions etc. to be closer to nautobots version.

pyproject.toml
```diff
@@ -58,6 +52,32 @@ toml = "^0.10.2"
 attrs = "^23.2.0"
 towncrier = "^24.8.0"
 ruff = "*"
+Markdown = "*"
+# Rendering docs to HTML
+mkdocs = "~1.6.1"
+# Embedding YAML files into Markdown documents as tables
+markdown-data-tables = "~1.0.0"
+# Render custom markdown for version added/changed/remove notes
+markdown-version-annotations = "~1.0.1"
```
itdependsnetworks: These should be pinned and the same as the requirements.txt. Ideally we can get this over too: https://github.com/nautobot/nautobot-app-golden-config/blob/develop/nautobot_golden_config/tests/test_basic.py

jeffkala: updated both.

itdependsnetworks (approved):

PR #338, opened 2025-07-21
Adding file_download and file_upload

itdependsnetworks:
creating a PR from this so I can comment.

### DIFF
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -730,6 +730,28 @@ class IOSDevice(BaseDevice):
             return True
         return self.verify_md5(filename, md5, file_system)
 
+    def file_upload(self, filename, url, file_system=None, read_timeout=2000):
+        """Upload file to remote server.
+
+        Examples:
+            # Upload file to remote server
+            >>> device.file_upload("image.bin", "scp://user:password@192.168.1.1/image.bin")
+            # Upload running config to remote server (override default file system)
+            >>> device.file_upload("running-config", "scp://user:password@192.168.1.1/config.txt", file_system="")
+
+        Args:
+            filename (str): Name of the file to upload.
+            url (str): URL of the file to upload to. (e.g., "scp://user:password@192.168.1.1/image.bin")
+            file_system (str, optional): File system the file is on.
+            read_timeout (int, optional): Netmiko timeout when waiting for device prompt. Default 2000.
+        """
+        self.enable()
+        if file_system is None:
+            file_system = self._get_file_system()
+
+        copy_command = f"copy {file_system}{filename} {url}"
+        self.show(copy_command, read_timeout=read_timeout)
+
     def file_md5(self, filename, file_system=None, read_timeout=1000):
         """Return the MD5 hash of a file on the device.
 
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -685,6 +685,80 @@ class IOSDevice(BaseDevice):
         log.debug("Host %s: File %s does not already exist on remote.", self.host, src)
         return False
 
+    def file_download(
+        self, protocol, src, path, username=None, password=None, dest="", file_system=None, read_timeout=2000
+    ):
+        """Download file from remote server.
+
+        Args:
+            protocol (str): Protocol to use for download (e.g., "scp", "ftp", "http", "https")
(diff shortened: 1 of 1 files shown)


pyntc/devices/ios_device.py
```diff
@@ -685,6 +685,80 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
         log.debug("Host %s: File %s does not already exist on remote.", self.host, src)
         return False
 
+    def file_download(
+        self, protocol, src, path, username=None, password=None, dest="", file_system=None, read_timeout=2000
```
itdependsnetworks: To me... we should just provide the url, it will be different for many different use cases, and the user can figure out how they want to handle usernames or not.

pyntc/devices/ios_device.py
```diff
@@ -685,6 +685,80 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
         log.debug("Host %s: File %s does not already exist on remote.", self.host, src)
         return False
 
+    def file_download(
```
itdependsnetworks: Is there anyway we can support idempotentcy? Even if we can't always support it, is there some ways to support it in some cases?

pyntc/devices/ios_device.py
joewesch: Changed

pyntc/devices/ios_device.py
joewesch: Added both filename idempotency as well as md5 verification to download.

pyntc/devices/ios_device.py
```diff
@@ -685,6 +685,109 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
         log.debug("Host %s: File %s does not already exist on remote.", self.host, src)
         return False
 
+    def file_download(self, url, dest="", md5=None, file_system=None, read_timeout=2000):
```
jeffkala, line 688: Should we just do `**netmiko_kwargs` too?  Incase you need `read_timeout_override` or any of the other netmiko kwargs.

pyntc/devices/ios_device.py
```diff
@@ -685,6 +685,109 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+                log.info("Host %s: File %s already exists on remote.", self.host, filename)
+                return True
+            if file_md5 != md5:
+                log.warning(
+                    "Host %s: File %s already exists on remote but MD5 hash mismatch. Downloading again.",
+                    self.host,
+                    filename,
+                )
+
+        copy_command = f"copy {url} {file_system}{dest}"
```
jeffkala, line 725: Wondering if we should construct this with some validation.  I'm thinking file_system would be `bootflash` the dest you'd have to know to do `///image.bin` or `//something/image.bin`. Can we use `Path` or some other lib to try and construct this more intelligently?

jeffkala: Even more so around the colon part.  Would expect `bootflash:///image.bin`

pyntc/devices/ios_device.py
```diff
@@ -685,6 +685,109 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+
+        self.show(copy_command, read_timeout=read_timeout)
+
+        if not md5:
+            return True
+        return self.verify_md5(filename, md5, file_system)
+
+    def file_upload(self, filename, url, file_system=None, read_timeout=2000):
+        """Upload file to remote server.
```
jeffkala, line 734:
```suggestion
        """Upload local file to remote server.
```

pyntc/devices/ios_device.py
```diff
@@ -685,6 +685,109 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+
+        copy_command = f"copy {url} {file_system}{dest}"
+
+        self.show(copy_command, read_timeout=read_timeout)
+
+        if not md5:
+            return True
+        return self.verify_md5(filename, md5, file_system)
+
+    def file_upload(self, filename, url, file_system=None, read_timeout=2000):
```
jeffkala, line 733: perhaps same as before around **netmiko_kwargs

jeffkala (changes requested):

joewesch:
We have decided to instead run these through nornir-nautobot via a dispatcher method instead, so I'm going to close this PR.

PR #344, opened 2026-02-17
updating re_show_redundancy to check for null value

cheenpo:
NTC-5197


cheenpo:
copying in feedback where this original commit was patched:
```
We did attempt your fix, and that did work to allow the file transfers to be conducted on routers.
That being said, we also noticed that the cause of the errors were caused by console logs bleeding through in the command output causing the re.match to anchor on those logs, so, we switched re.match to re.search and also rectified the issue 
```

jeffkala:
Needs unittest

gsnider2195:
> we also noticed that the cause of the errors were caused by console logs bleeding through in the command output

It sounds like `logging synchronous` may be turned off?

PR #345, opened 2026-02-18
Add file_pull option to Cisco IOS devices

smk4664:
This PR add the ability to have the Cisco IOS devices download the files instead of relying on copying the files to the device through scp.  I am using the dataclass FilePullSpec to simplify the required arguments when downloading the files instead of copying them through scp.

### DIFF
--- a/changes/345.added
+++ b/changes/345.added
@@ -1,0 +1,1 @@
+Added the ability to pull files from within a Cisco IOS device.
--- a/changes/345.added
+++ b/changes/345.added
@@ -1,1 +1,1 @@
-Added the ability to pull files from within a Cisco IOS device.
+Added the ability to download files from within a Cisco IOS device.
--- a/docs/user/lib_getting_started.md
+++ b/docs/user/lib_getting_started.md
@@ -272,7 +272,7 @@ from pyntc.utils.models import FileCopyModel
 >>>
 ```
 
-We recommend setting up the ip client on the device to be able to use this feature. For instance, on a Cisco IOS device you would need to set the source interface for the ip client when using http or https urls.  You can do this with the `config` method:
+Before using this feature you may need to configure a client on the device. For instance, on a Cisco IOS device you would need to set the source interface for the ip http client when using http or https urls.  You can do this with the `config` method:
 
 ```python
 >>> csr1.config('ip http client source-interface GigabitEthernet1')
--- a/docs/user/lib_getting_started.md
+++ b/docs/user/lib_getting_started.md
@@ -250,6 +250,33 @@ interface GigabitEthernet1
 >>>
 ```
 
+#### File Copy with URL
+
+All devices that support file copy also support copying files directly from a URL to the device.  This is useful for larger files like OS images.  To do this, you need to use the `FilePullSpec` data model to specify the source file information and then pass that to the `file_copy` method. Currently only supported on Cisco IOS devices. Tested with ftp, http, https, sftp, and tftp urls.
+
+```python
+from pyntc.utils.models import FilePullSpec
+
+>>> source_file = FilePullSpec(
+...     download_url='http://example.com/newconfig.cfg',
+...     checksum='abc123def456',
+...     hashing_algorithm='md5',
+...     file_name='newconfig.cfg'
+... )
+>>> for device in devices:
+...    device.file_copy(source_file)
+...
+>>>
+```
+
+We recommend setting up the ip client on the device to be able to use this feature. For instance, on a Cisco IOS device you would need to set the source interface for the ip client when using http or https urls.  You can do this with the `config` method:
+
+```python
+>>> csr1.config('ip http client source-interface GigabitEthernet1')
+>>>
+```
+
+
 ### Save Configs
 
 - `save` method
--- a/docs/user/lib_getting_started.md
+++ b/docs/user/lib_getting_started.md
@@ -250,21 +250,24 @@ interface GigabitEthernet1
 >>>
 ```
 
-#### File Copy with URL
+#### Remote File Copy (Download to Device)
 
-All devices that support file copy also support copying files directly from a URL to the device.  This is useful for larger files like OS images.  To do this, you need to use the `FilePullSpec` data model to specify the source file information and then pass that to the `file_copy` method. Currently only supported on Cisco IOS devices. Tested with ftp, http, https, sftp, and tftp urls.
+Some devices support copying files directly from a URL to the device. This is useful for larger files like OS images.  To do this, you need to use the `FileCopyModel` data model to specify the source file information and then pass that to the `remote_file_copy` method. Currently only supported on Cisco IOS devices. Tested with ftp, http, https, sftp, and tftp urls.
+
+- `remote_file_copy` method
 
 ```python
-from pyntc.utils.models import FilePullSpec
+from pyntc.utils.models import FileCopyModel
 
->>> source_file = FilePullSpec(
-...     download_url='http://example.com/newconfig.cfg',
+>>> source_file = FileCopyModel(
+...     download_url='sftp://example.com/newconfig.cfg',
 ...     checksum='abc123def456',
 ...     hashing_algorithm='md5',
-...     file_name='newconfig.cfg'
+...     file_name='newconfig.cfg',
+        vrf='Mgmt-vrf'
 ... )
 >>> for device in devices:
-...    device.file_copy(source_file)
+...    device.remote_file_copy(source_file)
 ...
 >>>
 ```
@@ -275,7 +278,6 @@ We recommend setting up the ip client on the device to be able to use this featu
 >>> csr1.config('ip http client source-interface GigabitEthernet1')
 >>>
 ```
-
 
 ### Save Configs
 
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -298,10 +298,7 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
             (str): The hex digest of the file.
         """
         # Initialize the hash object dynamically
-        try:
-            file_hash = hashlib.new(hashing_algorithm.lower())
-        except ValueError as e:
-            raise ValueError(f"Unsupported hashing algorithm: {hashing_algorithm}") from e
+        file_hash = hashlib.new(hashing_algorithm.lower())
 
         with open(filepath, "rb") as f:
             # Read in chunks to handle large firmware files without RAM spikes
@@ -321,6 +318,11 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
             filename (str): The name of the file to check for on the remote device.
             hashing_algorithm (str): The hashing algorithm to use (default: "md5").
             kwargs (dict): Additional keyword arguments that may be used by subclasses.
+
+        Keyword Args:
+            file_system (str): Supported only for IOS and NXOS. The file system for the
+                remote file. If no file_system is provided, then the ``get_file_system``
+                method is used to determine the correct file system to use.
 
         Returns:
             (bool): True if the checksums match, False otherwise.
@@ -351,12 +353,15 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
             hashing_algorithm (str): The hashing algorithm to use (default: "md5").
             kwargs (dict): Additional keyword arguments that may be used by subclasses.
 
+        Keyword Args:
+            file_system (str): Supported only for IOS and NXOS. The file system for the
+                remote file. If no file_system is provided, then the ``get_file_system``
+                method is used to determine the correct file system to use.
+
         Returns:
             (bool): True if the file is verified successfully, False otherwise.
         """
-        return self.check_file_exists(filename, **kwargs) and self.compare_file_checksum(
-            checksum, filename, hashing_algorithm, **kwargs
-        )
+        raise NotImplementedError
 
     def install_os(self, image_name, **vendor_specifics):
         """Install the OS from specified image_name.
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -221,7 +221,7 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
 
         Keyword Args:
             file_system (str): Supported only for IOS and NXOS. The file system for the
-                remote fle. If no file_system is provided, then the ``get_file_system``
+                remote file. If no file_system is provided, then the ``get_file_system``
                 method is used to determine the correct file system to use.
         """
         raise NotImplementedError
@@ -241,7 +241,7 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
 
         Keyword Args:
             file_system (str): Supported only for IOS and NXOS. The file system for the
-                remote fle. If no file_system is provided, then the ``get_file_system``
+                remote file. If no file_system is provided, then the ``get_file_system``
                 method is used to determine the correct file system to use.
 
         Returns:
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -1,9 +1,11 @@
 """The module contains the base class that all device classes must inherit from."""
 
+import hashlib
 import importlib
 import warnings
 
 from pyntc.errors import FeatureNotFoundError, NTCError
+from pyntc.utils.models import FileCopyModel
 
 
 def fix_docs(cls):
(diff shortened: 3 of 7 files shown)


changes/345.added
```diff
@@ -0,0 +1 @@
+Added the ability to pull files from within a Cisco IOS device.
```
jeffkala: maybe instead of pull, -> download?

jeffkala (approved):
seems good to me.

jeffkala (approved):

pyntc/devices/ios_device.py
```diff
@@ -3,13 +3,18 @@
...
+from netmiko.cisco import CiscoIosFileTransfer
 from netmiko.exceptions import ReadTimeout
 
-from pyntc import log
-from pyntc.devices.base_device import BaseDevice, RollbackError, fix_docs
-from pyntc.errors import (
+if TYPE_CHECKING:
+    from netmiko.base_connection import BaseConnection
+
+from pyntc import log  # pylint: disable=wrong-import-position
```
gsnider2195: Would it make pylint and isort happy if you moved the conditional import above down below all of the other imports?

pyntc/devices/ios_device.py
```diff
@@ -36,6 +42,129 @@
...
+        expect_regex = rf"(Destination filename|{re.escape(current_prompt)}|confirm|Password|Address or name of remote host|Source username|Source filename|yes/no|Are you sure you want to continue connecting)"
+        command = f"copy {url} {self.file_system}{self.dest_file}"
+        # Use VRF specified, unless using http or https, which don't support vrf in the copy command
+        if self.source_file.vrf and self.source_file.scheme not in ["http", "https"]:
+            command = f"{command} vrf {self.source_file.vrf}"
+        output = self.ssh_ctl_chan.send_command(command, expect_string=expect_regex, read_timeout=300)
+        for _ in range(10):
+            if current_prompt in output:
+                return
+            # Assume that the filename and address are sent with the url.
```
gsnider2195:
```suggestion
            # Accept defaults for these prompts because the filename and address are sent with the url.
```

pyntc/devices/ios_device.py
```diff
@@ -36,6 +42,129 @@
...
+    def pull_file(self) -> None:
+        """Download the file from the URL to the device using the device's own download capabilities."""
+        url = self.source_file.clean_url
+        current_prompt = self.ssh_ctl_chan.find_prompt()
+        expect_regex = rf"(Destination filename|{re.escape(current_prompt)}|confirm|Password|Address or name of remote host|Source username|Source filename|yes/no|Are you sure you want to continue connecting)"
+        command = f"copy {url} {self.file_system}{self.dest_file}"
+        # Use VRF specified, unless using http or https, which don't support vrf in the copy command
+        if self.source_file.vrf and self.source_file.scheme not in ["http", "https"]:
+            command = f"{command} vrf {self.source_file.vrf}"
+        output = self.ssh_ctl_chan.send_command(command, expect_string=expect_regex, read_timeout=300)
```
gsnider2195: Is this going to fail if the file transfer takes longer than 5 minutes? While the file is transferring netmiko is just waiting to match on one of the prompts right? I think we're still going to see use cases where large files are transferred over legacy circuits that may take a long time.

pyntc/devices/ios_device.py
```diff
@@ -36,6 +42,129 @@
...
+        for _ in range(10):
```
gsnider2195: I think this should be an endless loop or a `while current_prompt not in output` and we probably need to raise an exception if none of the prompts match?

pyntc/devices/ios_device.py
```diff
@@ -36,6 +42,129 @@
...
+            if current_prompt in output:
```
gsnider2195: There's a 1:1 mapping of prompts in this loop and prompts in `expect_regex` that's being passed to netmiko above but that's not clearly defined as a requirement. If someone adds another prompt to the regex above but forgets to add it here this would fall through and it would be difficult to troubleshoot. If netmiko doesn't already have a pattern for responding to prompts we should write our own here. eg:

```py
prompt_answers = {
    current_prompt: None,  # break out of the loop
    r"(confirm|Address or name of remote host|Source filename|Destination filename)": "",  # Press enter
    r"Password": self.source_file.token,
    r"Source username": self.source_file.username,
    r"yes/no|Are you sure you want to continue connecting": "yes",
}

Comment (gsnider2195) [reply pyntc/devices/ios_device.py]: We could build the `expect_regex` from `'|'.join(prompt_answers.keys())` so that the two code paths are always in sync.

File: pyntc/devices/ios_device.py [hunk 7]
```diff
@@ -36,6 +42,129 @@
...
+            if self.direction == "put":
+                remote_file = self.dest_file
+            elif self.direction == "get":
+                remote_file = self.source_file
+        remote_sha512_cmd = f"{base_cmd} {self.file_system}/{remote_file}"
+        dest_sha512 = self.ssh_ctl_chan._send_command_str(remote_sha512_cmd, read_timeout=300)  # pylint: disable=protected-access
+        dest_sha512 = self.process_md5(dest_sha512)  # Process MD5 still does what we want for parsing the output.
+        return dest_sha512
+
+    def compare_md5(self) -> bool:
```
gsnider2195: I think we're trying too hard to fit a square peg into a round hole subclassing the `CiscoIosFileTransfer` class. I would prefer if we just vendored the whole thing and rip out the parts we don't need (get/put). That gives us more flexibility to update it as we see fit to support things like different hashing algorithms. I know I won't be able to remember all of the workarounds that were required to make this work when the time comes to add support for another platform.

pyntc/utils/models.py
```diff
@@ -0,0 +1,73 @@
...
+
+@dataclass
+class FilePullSpec:
+    """Data class to represent the specification for pulling a file from a URL to a network device.
+
+    Args:
+        download_url (str): The URL to download the file from. Can include credentials, but it's recommended to use the username and token fields instead for security reasons.
+        checksum (str): The expected checksum of the file.
+        file_name (str): The name of the file to be saved on the device.
+        hashing_algorithm (str): The hashing algorithm to use for checksum verification. Defaults to "md5".
```
gsnider2195:
```suggestion
        hashing_algorithm (str, optional): The hashing algorithm to use for checksum verification. Defaults to "md5".
```

pyntc/devices/ios_device.py
smk4664: Fair, I didn't know what to set this to. I will increase and make this a variable.

pyntc/devices/ios_device.py
smk4664: That sounds like it is going to be the solution I will test that.

pyntc/devices/ios_device.py
```diff
@@ -653,17 +738,88 @@ def file_copy(self, src, dest=None, file_system=None):
...
+        """Copy a file to a remote device.
+
+        Args:
+            src (FileCopyModel): The source file model.
+            dest (str): The destination file path on the remote device.
+            kwargs (dict): Additional keyword arguments that may be used by subclasses.
+
+        Keyword Args:
+            file_system (str): Supported only for IOS and NXOS. The file system for the
+                remote file. If no file_system is provided, then the ``get_file_system``
```
mattmiller87, line 775: what does the double backtick indicate here? Can this be a single backtick?
> ``get_file_system``

pyntc/devices/ios_device.py
```diff
@@ -611,6 +613,85 @@ def config_register(self):
...
+                method is used to determine the correct file system to use.
+            kwargs (dict): Additional keyword arguments that may be used by subclasses.
+
+        Returns:
+            (bool): True if the remote file exists, False if it doesn't.
+
+        Raises:
+            CommandError: If there is an error in executing the command to check if the file exists.
+        """
+        cmd = f"dir {file_system or self._get_file_system()}/{filename}"
```
mattmiller87, line 676: I really like this check.

pyntc/devices/ios_device.py
```diff
@@ -653,17 +738,88 @@ def file_copy(self, src, dest=None, file_system=None):
...
             self.open()
 
-            if not self.file_copy_remote_exists(src, dest, file_system):
+            if not self.verify_file(local_checksum, dest, file_system=file_system):
                 log.error(
                     "Host %s: Attempted file copy, but could not validate file existed after transfer %s",
                     self.host,
                     FileTransferError.default_message,
                 )
                 raise FileTransferError
```
mattmiller87, line 763: I'm unsure how this works nested in an if statement performing the same action.  I feel like there could be a better way.

pyntc/devices/ios_device.py
```diff
@@ -653,17 +738,88 @@ def file_copy(self, src, dest=None, file_system=None):
...
+            raise TypeError("src must be an instance of FileCopyModel")
+        if file_system is None:
+            file_system = self._get_file_system()
+        if dest is None:
+            dest = src.file_name
+        if not self.verify_file(src.checksum, dest, hashing_algorithm=src.hashing_algorithm, file_system=file_system):
+            current_prompt = self.native.find_prompt()
+
+            # Define prompt mapping for expected prompts during file copy
+            prompt_answers = {
```
mattmiller87, line 792: 👏  great dictionary. really helps in your later code.

pyntc/devices/ios_device.py
```diff
@@ -653,17 +738,88 @@ def file_copy(self, src, dest=None, file_system=None):
...
+                            answer, expect_string=expect_regex, read_timeout=900, cmd_verify=not is_password
+                        )
+                        break  # Exit the for loop and check the new output for the next prompt
+
+        if not self.verify_file(src.checksum, dest, hashing_algorithm=src.hashing_algorithm, file_system=file_system):
+            log.error(
+                "Host %s: Attempted remote file copy, but could not validate file existed after transfer %s",
+                self.host,
+                FileTransferError.default_message,
+            )
```
mattmiller87: I noticed these above, but was curious for this log.error we're defining.  We're using a mix of f-strings and %s format. Does logging not like the use of f-strings?

Also applies to Lines: 796-8, 802
compared to non-logging format in Lines: 783, 787

pyntc/devices/ios_device.py
smk4664: This is how all the docstrings for pyntc are. I am not sure why, but I kept the same as existing.

docs/user/lib_getting_started.md
```diff
@@ -250,6 +250,35 @@ interface GigabitEthernet1
...
+...     file_name='newconfig.cfg',
+        vrf='Mgmt-vrf'
+... )
+>>> for device in devices:
+...    device.remote_file_copy(source_file)
+...
+>>>
+```
+
+We recommend setting up the ip client on the device to be able to use this feature. For instance, on a Cisco IOS device you would need to set the source interface for the ip client when using http or https urls.  You can do this with the `config` method:
```
gsnider2195:
```suggestion
Before using this feature you may need to configure a client on the device. For instance, on a Cisco IOS device you would need to set the source interface for the ip http client when using http or https urls.  You can do this with the `config` method:
```

pyntc/devices/ios_device.py
smk4664: So, I don't want the check to run twice if the file actually verified. But I do want to validate that the file was copied after the copy finishes. And sharing that, I now realize that I did not indent on the remote_file_copy. So, I am pushing a new commit to indent.

pyntc/devices/ios_device.py
smk4664: Correct, logging is not computed until the log is executed.  The recommendation is to use %s formatting for logs to save on the processing of f-strings. See https://docs.astral.sh/ruff/rules/logging-f-string/

pyntc/devices/base_device.py
```diff
@@ -241,13 +243,121 @@ def file_copy_remote_exists(self, src, dest=None, **kwargs):
...
+            add_newline (bool): Whether to append a newline before final hashing (Some devices may require this).
+
+        Returns:
+            (str): The hex digest of the file.
+        """
+        # Initialize the hash object dynamically
+        try:
+            file_hash = hashlib.new(hashing_algorithm.lower())
+        except ValueError as e:
+            raise ValueError(f"Unsupported hashing algorithm: {hashing_algorithm}") from e
```
gsnider2195: It's probably fine to just let the original exception raise?

```py
>>> hashlib.new("hello")
  File "/usr/lib/python3.12/hashlib.py", line 160, in __hash_new
    return _hashlib.new(name, data, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_hashlib.UnsupportedDigestmodError: [digital envelope routines] unsupported

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/hashlib.py", line 166, in __hash_new
    return __get_builtin_constructor(name)(data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/hashlib.py", line 123, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type hello
```

pyntc/devices/base_device.py
```diff
@@ -241,13 +243,121 @@ def file_copy_remote_exists(self, src, dest=None, **kwargs):
...
+        raise NotImplementedError
+
+    @staticmethod
+    def get_local_checksum(filepath, hashing_algorithm="md5", add_newline=False):
+        """Get the checksum of a local file using a specified algorithm.
+
+        Args:
+            filepath (str): The path to the local file.
+            hashing_algorithm (str): The hashing algorithm to use (e.g., "md5", "sha256").
+            add_newline (bool): Whether to append a newline before final hashing (Some devices may require this).
```
gsnider2195, line 295: Did this come from netmiko? Seems weird but if it's in netmiko it's probably fixing some strange quirk in a device

pyntc/devices/base_device.py
```diff
@@ -241,13 +243,121 @@ def file_copy_remote_exists(self, src, dest=None, **kwargs):
...
+
+        Keyword Args:
+            file_system (str): Supported only for IOS and NXOS. The file system for the
+                remote file. If no file_system is provided, then the ``get_file_system``
+                method is used to determine the correct file system to use.
+        """
+        raise NotImplementedError
+
+    def verify_file(self, checksum, filename, hashing_algorithm="md5", **kwargs):
+        """Verify a file on the remote device by and validate the checksums.
```
gsnider2195:
```suggestion
        """Verify a file on the remote device by confirming the file exists and validate the checksum.
```

pyntc/devices/base_device.py
```diff
@@ -241,13 +243,121 @@ def file_copy_remote_exists(self, src, dest=None, **kwargs):
...
+
+        Args:
+            checksum (str): The checksum of the file.
+            filename (str): The name of the file to check for on the remote device.
+            hashing_algorithm (str): The hashing algorithm to use (default: "md5").
+            kwargs (dict): Additional keyword arguments that may be used by subclasses.
```
gsnider2195, line 354: Does this need the `file_system` keyword arg docstring?

pyntc/devices/base_device.py
```diff
@@ -241,13 +243,121 @@ def file_copy_remote_exists(self, src, dest=None, **kwargs):
...
+        return file_hash.hexdigest()
+
+    def compare_file_checksum(self, checksum, filename, hashing_algorithm="md5", **kwargs):
+        """Compare the checksum of a local file with a remote file.
+
+        Args:
+            checksum (str): The checksum of the file.
+            filename (str): The name of the file to check for on the remote device.
+            hashing_algorithm (str): The hashing algorithm to use (default: "md5").
+            kwargs (dict): Additional keyword arguments that may be used by subclasses.
```
gsnider2195, line 320: Does this need the `file_system` keyword arg docstring?

pyntc/devices/base_device.py
```diff
@@ -241,13 +243,121 @@ def file_copy_remote_exists(self, src, dest=None, **kwargs):
...
+
+        Returns:
+            (bool): True if the file is verified successfully, False otherwise.
+        """
+        return self.check_file_exists(filename, **kwargs) and self.compare_file_checksum(
+            checksum, filename, hashing_algorithm, **kwargs
+        )
```
gsnider2195: It feels like we could get into trouble passing `**kwargs` into both of these methods. It could become a problem if someone overloads `check_file_exists` or `compare_file_checksum` and the kwargs don't match.

pyntc/devices/ios_device.py
```diff
@@ -611,6 +613,85 @@ def config_register(self):
         log.debug("Host %s: Config register %s", self.host, self._config_register)
         return self._config_register
 
+    def get_remote_checksum(self, filename, hashing_algorithm="md5", file_system=None, **kwargs):
+        """Get the checksum of a remote file.
+
+        Args:
+            filename (str): The name of the file to check for on the remote device.
+            hashing_algorithm (str): The hashing algorithm to use (default: "md5").
```
gsnider2195:
```suggestion
            hashing_algorithm (str): The hashing algorithm to use. Valid choices are "md5" and "sha512" (default: "md5").
```

pyntc/devices/ios_device.py
```diff
@@ -611,6 +613,85 @@ def config_register(self):
...
+            ValueError: If an unsupported hashing algorithm is provided.
+            CommandError: If there is an error in executing the command to get the remote checksum.
+        """
+        if hashing_algorithm not in {"md5", "sha512"}:
+            raise ValueError("hashing_algorithm must be either 'md5' or 'sha512' for Cisco IOS devices.")
+        if file_system is None:
+            file_system = self._get_file_system()
+        cmd = f"verify /{hashing_algorithm} {file_system}/{filename}"
+        result = self.native.send_command_timing(cmd, read_timeout=300)
+        if match := re.search(r"=\s+(\S+)", result):
```
gsnider2195: We may need to come back to this later and make this regex more robust depending on how various different ios/iosxe platforms and versions output this command.

pyntc/devices/ios_device.py
```diff
@@ -611,6 +613,85 @@ def config_register(self):
...
+
+        Returns:
+            (bool): True if the remote file exists, False if it doesn't.
+
+        Raises:
+            CommandError: If there is an error in executing the command to check if the file exists.
+        """
+        cmd = f"dir {file_system or self._get_file_system()}/{filename}"
+        result = self.native.send_command(cmd, read_timeout=30)
+        print(f"Result: {result}")
```
gsnider2195: remove?

pyntc/devices/ios_device.py
```diff
@@ -611,6 +613,85 @@ def config_register(self):
...
+        log.info(
```
gsnider2195: Should this be a debug log?

pyntc/devices/ios_device.py
```diff
@@ -611,6 +613,85 @@ def config_register(self):
...
+            cmd,
+            result,
+        )
+        if re.search(r"No such file|No files found|Path does not exist|Error opening", result):
+            log.debug("Host %s: File %s does not exist on remote.", self.host, filename)
+            return False
+        if re.search(rf"Directory of .*{filename}", result):
+            log.debug("Host %s: File %s exists on remote.", self.host, filename)
+            return True
+        raise CommandError(cmd, f"Unable to determine if file {filename} exists on remote.")
```
gsnider2195: This should probably contain the result?

pyntc/devices/ios_device.py
```diff
@@ -653,17 +738,90 @@ def file_copy(self, src, dest=None, file_system=None):
...
+            }
+            keys = list(prompt_answers.keys()) + [re.escape(current_prompt)]
+            expect_regex = f"({'|'.join(keys)})"
+
+            command = f"copy {src.clean_url} {file_system}{dest}"
+            if src.vrf and src.scheme not in {"http", "https"}:
+                command = f"{command} vrf {src.vrf}"
+
+            # _send_command currently checks for % and raises an error, but during the file copy
+            # their may be a % warning that does not indicate a failure so we will use send_command directly.
```
gsnider2195:
```suggestion
            # there may be a % warning that does not indicate a failure so we will use send_command directly.
```

gsnider2195: Great comment btw

pyntc/devices/ios_device.py
```diff
@@ -653,17 +738,90 @@ def file_copy(self, src, dest=None, file_system=None):
...
+            output = self.native.send_command(command, expect_string=expect_regex, read_timeout=900)
```
gsnider2195: Do we need to make the timeout configurable for large files/slow connections?

pyntc/devices/base_device.py
smk4664: It is in Netmiko, which is where it came from, but doesn't seem to be used for IOS. Probably fixing something in some devices.

pyntc/devices/base_device.py
smk4664: Looks like it should. I will add.

pyntc/devices/ios_device.py
smk4664: Thanks! I left this in while debugging the test and forgot to remove.

pyntc/devices/base_device.py
smk4664: This is the pattern that the maintainers took in order to pass `file_system` when dealing with IOS devices. Is there a better way to do this?

pyntc/devices/ios_device.py
smk4664: Yes, looking at Netmiko (where I got this regex from) IOS XR has a very different output. Let me try to support both, since pyntc does not differentiate between IOS and IOSXR

pyntc/devices/ios_device.py
smk4664: Yes, changing

pyntc/devices/ios_device.py
smk4664: Sure, I will add that, Netmiko did not, but that is likely due to the way scp works.

pyntc/devices/ios_device.py
gsnider2195: It seems we just don't support IOS-XR because they're not similar enough for the IOS modules to work for XR.

gsnider2195 (approved):

jeffkala (approved):

PR #347, opened 2026-03-03
fix docs test failing after new model adds

jeffkala:


smk4664 (approved):
This looks great! Thanks for making this fix!

PR #355, opened 2026-03-07
fixes #354 github release failure

### LINKED ISSUE #354: Fix Github release failing due to bad credentials
### Environment
* Python version:  3.10
* pyntc version:  2.2.0

### Expected Behavior
release.yml properly publishes release to pypi and github

### Observed Behavior
Only pypi release is working

### Steps to Reproduce
1. Do a release
2. Notice github release is failing with bad credentials.

jeffkala:
fixes #354


smk4664 (approved):
Thanks for the quick and correct fix.  Github creates its own token, so there was no need to use a different token.

PR #356, opened 2026-03-13
bump versions

icovada:
scp 0.14.5->0.15


jeffkala:
@icovada can you provide a short description on why here?  What does this give us.

icovada:
Currently, this is the only dependency in the entire cloud image that prevents us to upgrade to 0.15.

I sent this because it looks like everyone has already upgraded to the newest version safely

jeffkala:
@mattmiller87 @jtdub this affect any of your current work?

mattmiller87:
tldr; This should be fine.

The bulk of the change is removing the `^` and replacing with better min/max limits for the package. 

The biggest change is moving SCP to a new major version, `0.15`.  This impacts pyntc/devices/jnpr_device.py methods:
- file_copy
- rollback
- save

From what I can tell, none of this will be a breaking change. SCP doesn't document in their releases, so I was able to find some details in their changelog - https://github.com/jbardin/scp.py/blob/master/CHANGELOG.md

jtdub (approved):
LGTM

jeffkala (approved):

PR #365, opened 2026-04-10
Add Arista EOS remote file copy

mattmiller87:
## New Pull Request

Have you:
- [x] Updated the docs if necessary?
- [ ] Updated any configuration settings?
- [x] Written a unit test?

## Change Notes
Added the remote file copy feature to Arista EOS devices.
Added unittests for remote file copy on Arista EOS devices. (with help from Claude)

## Justification
This allows an Arista device to run the copy command directly.


docs/user/lib_getting_started.md
```diff
@@ -252,7 +252,7 @@ interface GigabitEthernet1
 
 #### Remote File Copy (Download to Device)
 
-Some devices support copying files directly from a URL to the device. This is useful for larger files like OS images.  To do this, you need to use the `FileCopyModel` data model to specify the source file information and then pass that to the `remote_file_copy` method. Currently only supported on Cisco IOS and Juniper Junos devices. Tested with ftp, http, https, sftp, and tftp urls.
+Some devices support copying files directly from a URL to the device. This is useful for larger files like OS images.  To do this, you need to use the `FileCopyModel` data model to specify the source file information and then pass that to the `remote_file_copy` method. Currently only supported on Cisco IOS, Juniper Junos, and Arista EOS devices. Tested with ftp, http, https, sftp, and tftp urls.
```
jeffkala:
```suggestion
Some devices support copying files directly from a URL to the device. This is useful for large files like OS images.  To do this, you need to use the `FileCopyModel` data model to specify the source file information and then pass that to the `remote_file_copy` method.  The model is currently supported on Cisco IOS, Juniper Junos, and Arista EOS devices. It has been tested with ftp, http, https, sftp, and tftp urls.
```

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,283 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+        # Normalize file_system to Arista format (e.g., "flash:" or "/mnt/flash")
+        if file_system.endswith(":"):
+            # Already in shorthand format like "flash:"
+            pass
+        elif file_system.startswith("/"):
+            # Full path format like "/mnt/flash" - keep as is
+            pass
+        else:
+            # Assume shorthand without colon (e.g., "flash" -> "flash:")
+            file_system = f"{file_system}:"
```
jeffkala, line 490:
```python
if not file_system.startswith("/") and not file_system.endswith(":"):
    file_system = f"{file_system}:"
```

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,283 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+        command, detect_prompt = command_builders[builder_key]()
+        log.debug("Host %s: Preparing copy command for %s", self.host, src.scheme)
+
+        # Execute copy command
+        if detect_prompt and src.token:
+            # Use send_command_timing for interactive password prompt
+            output = self.native_ssh.send_command_timing(command, read_timeout=src.timeout, cmd_verify=False)
+            log.debug("Host %s: Copy command (with timing) output: %s", self.host, output)
+
+            if "Password:" in output or "password:" in output:
```
jeffkala: if "password:" in output.lower():

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,283 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+                self.native_ssh.write_channel(src.token + "\n")
+                # Read the response after sending password
+                output += self.native_ssh.read_channel()
+                log.debug("Host %s: Output after password entry: %s", self.host, output)
+            elif any(error in output for error in ["Error", "Invalid", "Failed"]):
```
jeffkala: should we do output.lower()? and lower the values to perhaps catch any oddities in case sensitivity.

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,283 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+                log.error("Host %s: Error detected in copy command output: %s", self.host, output)
+                raise FileTransferError(f"Error detected in copy command output: {output}")
+        else:
+            # Use regular send_command for non-interactive transfers
+            output = self.native_ssh.send_command(command, read_timeout=src.timeout)
+            log.debug("Host %s: Copy command output: %s", self.host, output)
+
+            if any(error in output for error in ["Error", "Invalid", "Failed"]):
```
jeffkala: same a previous.

mattmiller87:
@jeffkala, I appreciate the comments. I've applied your suggestions.

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+        else:
+            raise CommandError(command, f"Unable to determine if file {filename} exists on remote: {result}")
+
+        return exists
+
+    def get_remote_checksum(self, filename, hashing_algorithm="md5", **kwargs):
+        """Get the checksum of a remote file on Arista EOS device using netmiko SSH.
+
+        Uses Arista's 'verify' command via SSH to compute file checksums.
+        Note, Netmiko FileTransfer only supports `verify /md5`
```
gsnider2195, line 471: What does this mean?

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+            log.debug(
+                "Host %s: Verify command '%s' returned: %s",
+                self.host,
+                command,
+                result,
+            )
+
+            # Parse the checksum from the output
+            # Expected format: verify /sha512 (flash:nautobot.png) = <checksum>
+            match = re.search(r"=\s*([a-fA-F0-9]+)", result)
```
gsnider2195, line 514: I don't like how imprecise this regex is. Can we just do `if remote_checksum.lower() in result.lower():`?

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+
+        # Build the path
+        if file_system.endswith(":"):
+            path = f"{file_system}{filename}"
+        else:
+            path = f"{file_system}/{filename}"
+
+        # Use Arista's verify command to get the checksum
+        # Example: verify /sha512 flash:nautobot.png
+        command = f"verify /{hashing_algorithm} {path}"
```
gsnider2195, line 500: We should handle invalid hashing_algorithm values and raise the correct exception so the user knows that the checksum type is invalid instead of just saying that it failed to parse. If they select SHA in their SoftwareImageFile they may never understand that their checksum type is wrong and will just assume that the file transfer is failing since `% Invalid input` is the only message they'll see:

```
Arista7050TX-64-R#verify /asdf flash:/EOS-4.15.4F.swi
% Invalid input
```

jeffkala (approved):
I'm good with it, thanks for updating my first set of reviews.

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+            log.error("Host %s: Could not parse checksum from verify output: %s", self.host, result)
+            raise CommandError(command, f"Could not parse checksum from verify output: {result}")
+
+        except Exception as e:
+            log.error("Host %s: Error getting remote checksum: %s", self.host, str(e))
+            raise CommandError(command, f"Error getting remote checksum: {str(e)}")
+
+    def _build_url_copy_command_simple(self, src, file_system):
+        """Build copy command for simple URL-based transfers (TFTP, HTTP, HTTPS without credentials)."""
+        return f"copy {src.download_url} {file_system}", False
```
gsnider2195, line 529: This should include the filename in the destination or it will default to the filename in the url which may not be correct. For example the filename in sftpgo's web server defaults to download because the url is `http://10.1.100.220:8080/web/client/pubshares/QVuvQMzugSegd2e7krDwr7/download`

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+
+    def _build_url_copy_command_with_creds(self, src, file_system):
+        """Build copy command for URL-based transfers with credentials (HTTP/HTTPS/SCP/FTP/SFTP)."""
+        parsed = urlparse(src.download_url)
```
gsnider2195, line 533: We probably need to fail if `parsed.query` is not blank since a url with a query contains a `?` which doesn't work in the EOS command line.

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+        port_str = f":{port}" if port else ""
+
+        # For HTTP/HTTPS, include both username and token
+        if src.scheme in ["http", "https"]:
+            command = f"copy {src.scheme}://{src.username}:{src.token}@{hostname}{port_str}{path} {file_system}"
+            detect_prompt = False
+        # For SCP/FTP/SFTP, include only username (password via prompt)
+        else:
+            command = f"copy {src.scheme}://{src.username}@{hostname}{port_str}{path} {file_system}"
+            detect_prompt = True
```
gsnider2195, line 556: We can clean this up a bit by using `netloc` and dropping the default port for `https` and `http`
```suggestion
        # For HTTP/HTTPS, include both username and token
        if src.scheme in ["http", "https"]:
            command = f"copy {src.scheme}://{src.username}:{src.token}@{netloc}{path} {file_system}"
            detect_prompt = False
        # For SCP/FTP/SFTP, include only username (password via prompt)
        else:
            command = f"copy {src.scheme}://{src.username}@{netloc}{path} {file_system}"
            detect_prompt = True
```

Here are some examples of netloc:

```py
>>> urlparse("http://10.1.100.220:8080/web/client/pubshares/QVuvQMzugSegd2e7krDwr7/download?compress=false").hostname
'10.1.100.220'
>>> urlparse("http://10.1.100.220:8080/web/client/pubshares/QVuvQMzugSegd2e7krDwr7/download?compress=false").netloc
'10.1.100.220:8080'
>>> urlparse("http://10.1.100.220/web/client/pubshares/QVuvQMzugSegd2e7krDwr7/download?compress=false").netloc
'10.1.100.220'
>>> urlparse("ftp://10.1.100.220:8443/web/client/pubshares/QVuvQMzugSegd2e7krDwr7/download?compress=false").netloc
'10.1.100.220:8443'
>>> urlparse("https://10.1.100.220:443/web/client/pubshares/QVuvQMzugSegd2e7krDwr7/download?compress=false").netloc
'10.1.100.220:443'
```

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+        log.debug("Host %s: Starting remote file copy for %s to %s/%s", self.host, src.file_name, file_system, dest)
+
+        # Open SSH connection and enable
+        self.open()
+        self.enable()
+
+        # Validate scheme
+        supported_schemes = ["http", "https", "scp", "ftp", "sftp", "tftp"]
+        if src.scheme not in supported_schemes:
+            raise ValueError(f"Unsupported scheme: {src.scheme}")
```
gsnider2195, line 596: Validate the scheme before trying to connect to the device. It will fail faster.

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+        return command, detect_prompt
+
+    def remote_file_copy(self, src: FileCopyModel, dest=None, file_system=None, include_username=False, **kwargs):
+        """Copy a file from remote source to device.
+
+        Args:
+            src (FileCopyModel): The source file model with transfer parameters.
+            dest (str): Destination filename (defaults to src.file_name).
+            file_system (str): Device filesystem (auto-detected if not provided).
+            include_username (bool): Whether to include username in the copy command. Defaults to False.
```
gsnider2195, line 567: Does `include_username` make sense here? Shouldn't we always include the username if it's provided and let the device prompt for the password?

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+        self.open()
+        self.enable()
+
+        # Validate scheme
+        supported_schemes = ["http", "https", "scp", "ftp", "sftp", "tftp"]
+        if src.scheme not in supported_schemes:
+            raise ValueError(f"Unsupported scheme: {src.scheme}")
+
+        # Build command based on scheme and credentials
+        command_builders = {
```
gsnider2195, line 599: If `include_username` is not required we would probably want to do something like this:

```py
if src.scheme == "tftp" or src.username is None:
    command, detect_prompt = self._build_url_copy_command_simple(src, file_system)
else:
    command, detect_prompt = self._build_url_copy_command_with_creds(src, file_system)
```

If they didn't provide credentials and the copy fails, we'll bubble up that failure to the user.

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+            **kwargs (Any): Passible parameters such as file_system.
+
+        Returns:
+            (bool): True if the file is verified successfully, False otherwise.
+        """
+        exists = self.check_file_exists(filename, **kwargs)
+        device_checksum = (
+            self.get_remote_checksum(filename, hashing_algorithm=hashing_algorithm, **kwargs) if exists else None
+        )
+        if checksum == device_checksum:
```
gsnider2195, line 683:
```suggestion
        if checksum.lower() == device_checksum.lower():
```

pyntc/devices/eos_device.py
```diff
@@ -421,6 +423,276 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+        Args:
+            checksum (str): The checksum of the file.
+            filename (str): The name of the file to check for on the remote device.
+            hashing_algorithm (str): The hashing algorithm to use (default: "md5").
+            **kwargs (Any): Passible parameters such as file_system.
+
+        Returns:
+            (bool): True if the file is verified successfully, False otherwise.
+        """
+        exists = self.check_file_exists(filename, **kwargs)
```
gsnider2195, line 679:
```suggestion
        if not self.check_file_exists(filename, **kwargs):
            log.debug("...")  # File not found log message here
            return False
```

tests/unit/test_devices/test_eos_device.py
```diff
@@ -433,3 +436,997 @@ def test_init_pass_port_and_timeout(mock_eos_connect):
...
+    method should raise a `TypeError`.
+
+    Validates: Requirements 1.2, 15.1
+    """
+    device = EOSDevice("host", "user", "pass")
+
+    with pytest.raises(TypeError) as exc_info:
+        device.remote_file_copy(src)
+
+    assert "src must be an instance of FileCopyModel" in str(exc_info.value)
```
gsnider2195, line 481: 99% of the apps ecosystem is unittest and `self.assertRaises`. Lets switch this over to those patterns for consistency.

tests/unit/test_devices/test_eos_device.py
```diff
@@ -433,3 +436,997 @@ def test_init_pass_port_and_timeout(mock_eos_connect):
...
+    class _ST:
+        @staticmethod
+        def just(value):
+            return value
+
+        @staticmethod
+        def one_of(*args):
+            return args[0]
+
+    st = _ST()
```
gsnider2195, line 462: We should stick with patterns that the rest of the team maintaining this is familiar with.

pyntc/devices/eos_device.py
gsnider2195: May also need to account for `src.username is None and src.token is not None` for URLs like https://token@example.com/file.bin

pyntc/devices/eos_device.py
mattmiller87: I was hoping the netmiko FileTransfer would solve this problem, but there is an issue where it is limited to md5. 

https://ktbyers.github.io/netmiko/docs/netmiko/arista/index.html > Expand Source Code
```
def remote_md5(
        self, base_cmd: str = "verify /md5", remote_file: Optional[str] = None
    ) -> str:
        if remote_file is None:
            if self.direction == "put":
                remote_file = self.dest_file
            elif self.direction == "get":
                remote_file = self.source_file
        remote_md5_cmd = f"{base_cmd} file:{self.file_system}/{remote_file}"
        dest_md5 = self.ssh_ctl_chan._send_command_str(remote_md5_cmd, read_timeout=600)
        dest_md5 = self.process_md5(dest_md5)
        return dest_md5
```

PR #368, opened 2026-04-14
Updates to the EOS driver based on PR comments from #365

jtdub:
## Summary

Addresses review comments from @gsnider2195 on PR #365. Resolves 12 items across `eos_device.py` and `test_eos_device.py`.

### Production code (`pyntc/devices/eos_device.py`)

- **Fail-fast validation**: Scheme and query string checks moved before `open()`/`enable()` so invalid inputs error without connecting to the device.
- **EOS hashing algorithm validation**: `get_remote_checksum` now validates against an EOS-specific subset (`md5`, `sha1`, `sha256`, `sha512`) and raises a clear `ValueError` for unsupported algorithms.
- **Checksum parsing simplified**: Replaced regex with `str.split("=")` for parsing `verify` command output.
- **Destination filename in copy commands**: Both `_build_url_copy_command_simple` and `_build_url_copy_command_with_creds` now include the destination filename so URL-derived names (e.g., `download`) don't override the intended filename.
- **Query string rejection**: URLs with `?` are rejected since EOS CLI cannot handle them.
- **`netloc` for host:port**: Replaced manual hostname/port construction with computed `netloc`.
- **Credential routing simplified**: Removed `include_username` parameter. Routing is now based on `src.username`: if `None`, uses the simple builder; otherwise uses the creds builder. Token-only auth (no username) correctly falls through to the simple builder.
- **`clean_url` instead of `download_url`**: Command builders use `src.clean_url` to prevent credential leakage.
- **`verify_file` improvements**: Early return when file doesn't exist; case-insensitive checksum comparison.
- **Error detection extracted**: Duplicate error-checking blocks consolidated into `_check_copy_output_for_errors`.
- **Module-level constants**: `EOS_SUPPORTED_SCHEMES` and `EOS_SUPPORTED_HASHING_ALGORITHMS` are now sets at module scope.

### Tests (`tests/unit/test_devices/test_eos_device.py`)

- Removed `hypothesis` import block and all standalone `@given`/`@pytest.mark.parametrize` test functions.
- Consolidated into `unittest.TestCase` classes with `self.assertRaises`, `self.assertEqual`, and `self.subTest`.
- Removed duplicate `TestRemoteFileCopyCommandExecution` class (tests already covered by `TestRemoteFileCopy`).
- Removed stale `# TODO: unit test for remote_file_copy`.
- Added tests for unsupported scheme, query string rejection, and token-only auth routing.

## Test plan

- [x] `poetry run python -m pytest tests/unit/test_devices/test_eos_device.py` — 65 passed, 19 subtests passed
- [x] `poetry run ruff check` — all checks passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

```
% source .env && poetry run pytest tests/integration/test_eos_device.py -v
===================================================================================== test session starts =====================================================================================
platform darwin -- Python 3.12.11, pytest-9.0.3, pluggy-1.6.0 -- /Users/jameswilliams/Library/Caches/pypoetry/virtualenvs/pyntc-gs002iVE-py3.12/bin/python
cachedir: .pytest_cache
rootdir: /Users/jameswilliams/Documents/code/pyntc
configfile: pyproject.toml
plugins: f5-sdk-3.0.21, requests-mock-1.12.1
collected 10 items

tests/integration/test_eos_device.py::test_device_connects PASSED                                                                                                                       [ 10%]
tests/integration/test_eos_device.py::test_check_file_exists_false PASSED                                                                                                               [ 20%]
tests/integration/test_eos_device.py::test_get_remote_checksum_after_exists PASSED                                                                                                      [ 30%]
tests/integration/test_eos_device.py::test_remote_file_copy_ftp PASSED                                                                                                                  [ 40%]
tests/integration/test_eos_device.py::test_remote_file_copy_tftp PASSED                                                                                                                 [ 50%]
tests/integration/test_eos_device.py::test_remote_file_copy_scp PASSED                                                                                                                  [ 60%]
tests/integration/test_eos_device.py::test_remote_file_copy_http PASSED                                                                                                                 [ 70%]
tests/integration/test_eos_device.py::test_remote_file_copy_https FAILED                                                                                                                [ 80%]
tests/integration/test_eos_device.py::test_remote_file_copy_sftp PASSED                                                                                                                 [ 90%]
tests/integration/test_eos_device.py::test_verify_file_after_copy PASSED                                                                                                                [100%]
```

The EOS device I'm testing against doesn't support the TLS version required by the server.

```
FAILED tests/integration/test_eos_device.py::test_remote_file_copy_https - pyntc.errors.FileTransferError: FileTransferError:
Error detected in copy command output:
cd `https://192.0.2.1/' [Waiting for response...]

% Error copying https://192.0.2.1/nautobot.png to flash: (
---- Connecting to 192.0.2.1 (192.0.2.1) port 8443
**** send: gnutls_handshake: A TLS fatal alert has been received.
**** recv: gnutls_handshake: A TLS fatal alert has been received.
cd: Fatal error: gnutls_handshake: A TLS fatal alert has been received.

see above for details
)
```

### DIFF
--- a/pyntc/devices/eos_device.py
+++ b/pyntc/devices/eos_device.py
@@ -26,6 +26,7 @@ from pyntc.errors import (
 from pyntc.utils import convert_list_by_key
 from pyntc.utils.models import FileCopyModel
 
+EOS_SUPPORTED_HASHING_ALGORITHMS = {"md5", "sha1", "sha256", "sha512"}
 BASIC_FACTS_KM = {"model": "modelName", "os_version": "internalVersion", "serial_number": "serialNumber"}
 INTERFACES_KM = {
     "speed": "bandwidth",
@@ -481,6 +482,13 @@ class EOSDevice(BaseDevice):
         Raises:
             CommandError: If the verify command fails (but not if file doesn't exist).
         """
+        # Validate hashing algorithm against EOS-supported values
+        if hashing_algorithm.lower() not in EOS_SUPPORTED_HASHING_ALGORITHMS:
+            raise ValueError(
+                f"Unsupported hashing algorithm '{hashing_algorithm}' for EOS. "
+                f"Supported algorithms: {sorted(EOS_SUPPORTED_HASHING_ALGORITHMS)}"
+            )
+
         file_system = kwargs.get("file_system")
         if file_system is None:
             file_system = self._get_file_system()
@@ -511,11 +519,11 @@ class EOSDevice(BaseDevice):
 
             # Parse the checksum from the output
             # Expected format: verify /sha512 (flash:nautobot.png) = <checksum>
-            match = re.search(r"=\s*([a-fA-F0-9]+)", result)
-            if match:
-                remote_checksum = match.group(1).lower()
-                log.debug("Host %s: Remote checksum for %s: %s", self.host, filename, remote_checksum)
-                return remote_checksum
+            if "=" in result:
+                remote_checksum = result.split("=")[-1].strip().lower()
+                if remote_checksum:
+                    log.debug("Host %s: Remote checksum for %s: %s", self.host, filename, remote_checksum)
+                    return remote_checksum
 
             log.error("Host %s: Could not parse checksum from verify output: %s", self.host, result)
             raise CommandError(command, f"Could not parse checksum from verify output: {result}")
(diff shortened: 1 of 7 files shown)


pyntc/devices/eos_device.py
```diff
@@ -524,123 +532,89 @@ def get_remote_checksum(self, filename, hashing_algorithm="md5", **kwargs):
...
+            # SCP/FTP/SFTP — password provided at the interactive prompt
+            command = f"copy {src.scheme}://{src.username}@{netloc}{path} {file_system}/{dest}"
             detect_prompt = True
 
         return command, detect_prompt
 
-    def remote_file_copy(self, src: FileCopyModel, dest=None, file_system=None, include_username=False, **kwargs):
+    def _check_copy_output_for_errors(self, output):
+        """Raise FileTransferError if copy command output contains error indicators."""
+        if any(error in output.lower() for error in ["error", "invalid", "failed"]):
```
jeffkala, line 569: am I crazy to think `if output.lower() in ["error", "invalid", "failed"]` is the same and much easier to read?

pyntc/devices/eos_device.py
```diff
@@ -524,123 +532,89 @@ def get_remote_checksum(self, filename, hashing_algorithm="md5", **kwargs):
...
+            log.error("Host %s: Error detected in copy command output: %s", self.host, output)
+            raise FileTransferError(f"Error detected in copy command output: {output}")
+
+    def remote_file_copy(self, src: FileCopyModel, dest=None, file_system=None, **kwargs):
```
jeffkala: The half baked typing drives me crazy, but not saying we have to hold up the PR for it.  Either do it or don't.

jeffkala (commented):
2 comments, but seems good for me

jtdub:
Responding to @jeffkala's comments:

---

**Re: `_check_copy_output_for_errors` readability** ([comment](https://github.com/networktocode/pyntc/pull/368#discussion_r3080941702))

Good callout on readability, but the logic isn't quite the same — `if output.lower() in ["error", "invalid", "failed"]` checks if the *entire* output string equals one of those words. What we need is substring matching: does the output *contain* any of those words. For example, `"% Error copying ftp://..."` would not match with `in`, but does match with the `any()` approach.

---

**Re: half-baked typing** ([comment](https://github.com/networktocode/pyntc/pull/368#discussion_r3080948736))

Fair point. Adding type hints to the `remote_file_copy` signature in the next push.

smk4664 (approved):
Thanks for fixing those issues I pointed out!

PR #370, opened 2026-04-20
Add EOS pre-transfer free-space check (NAPPS-1091)

jtdub:
## Summary

- Adds a pre-transfer free-space check to EOS `file_copy` (always on; uses `os.path.getsize` on the local source) and `remote_file_copy` (runs only when `FileCopyModel.file_size` is supplied — see "Fail-open semantics" below).
- Establishes a cross-driver seam on `BaseDevice` (`_get_free_space` per-platform probe, `_check_free_space` comparator, `_pre_transfer_space_check` fail-open wrapper) so companion work for IOS, NXOS, JUNOS, and ASA only has to implement the one platform-specific probe.
- Keeps `FileCopyModel.file_size` optional; adds a `file_size_unit` (`bytes` / `megabytes` / `gigabytes`; default `bytes`, always validated and lowercased) plus a computed `file_size_bytes` so callers can express size in human units and pyntc compares in raw bytes.

## Fail-open semantics

- If `FileCopyModel.file_size` is supplied, `remote_file_copy` uses it for the pre-transfer comparison and raises `NotEnoughFreeSpaceError` when the device lacks room. pyntc trusts the caller implicitly — no cross-check, no auto-discovery.
- If `FileCopyModel.file_size` is `None`, the pre-transfer check is skipped entirely. The transfer proceeds; a debug-level log records the skip. pyntc does not probe remote URLs for size — callers (a Nautobot OS Upgrades Job, an SSoT sync, or any pre-transfer script) populate `file_size` from their own metadata before handing the model to pyntc.
- Space-check failures raise `NotEnoughFreeSpaceError` specifically; they are never surfaced as checksum failures.

## Implementation notes

- `NotEnoughFreeSpaceError` gained optional `required` / `available` / `file_system` kwargs for a richer message; the positional `(hostname, min_space)` signature is preserved so `F5Device._check_free_space` keeps working untouched.
- EOS `_get_free_space` parses the trailing `(N bytes free)` token from `dir` output; the `file_system` argument is accepted for API parity but unused because EOS only exposes a single practical filesystem.
- `remote_file_copy` calls `_pre_transfer_space_check(src, file_system)` after scheme/URL validation and enable-session setup but before the copy command is built or sent — so an oversized transfer never reaches the device's copy machinery.
- `file_copy` compares `os.path.getsize(src)` against `_get_free_space()` after `file_copy_remote_exists()` reports the file is missing.
- `FileCopyModel.file_size_unit` is validated against `{bytes, megabytes, gigabytes}` (and lowercased) on every construction, regardless of whether `file_size` is supplied. The byte conversion (`file_size_bytes`) is gated on `file_size` being present.
- Integration-test scaffolding (`PROTOCOL_URL_VARS`, `build_file_copy_model`, `first_available_url`, `any_file_copy_model` fixture) lives in `tests/integration/_helpers.py` + `tests/integration/conftest.py` so EOS + ASA suites share one source of truth. Unit tests share an `EOSDeviceMockedTestCase` base class for the mocked-pyeapi setup.

## Not in this PR

- Auto-discovery probes inside pyntc. Callers populate `file_size` externally.
- IOS / NXOS / JUNOS / ASA implementations — follow-up tickets; each one now only needs to implement `_get_free_space` for its platform.

## Test plan

- [x] `poetry run pytest tests/unit -q` — **654 passed** (639 on develop; 12 new device tests + 3 new model tests)
- [x] `poetry run ruff check pyntc tests`
- [x] `poetry run ruff format --check pyntc tests`
- [x] `poetry run pylint pyntc` — **10.00/10**
- [x] `poetry run towncrier build --draft` renders `changes/370.added` as two bullets under `### Added`
- [x] Manual integration run against a lab cEOS node (`tests/integration/test_eos_device.py`) — covers `_get_free_space` / `_check_free_space` happy/raise paths, unit conversion end-to-end, oversized transfer rejected before the copy command, correctly-sized transfer completing, and the fail-open path (spy asserts `_check_free_space` is never called and the transfer still lands on flash).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -2,10 +2,13 @@
 
 import hashlib
 import importlib
+import logging
 import warnings
 
-from pyntc.errors import FeatureNotFoundError, NTCError
+from pyntc.errors import FeatureNotFoundError, NotEnoughFreeSpaceError, NTCError
 from pyntc.utils.models import FileCopyModel
+
+log = logging.getLogger(__name__)
 
 
 def fix_docs(cls):
@@ -51,6 +54,90 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
         self._serial_number = None
         self._model = None
         self._vlans = None
+
+    def _check_free_space(self, required_bytes, file_system=None):
+        """Raise NotEnoughFreeSpaceError when the target filesystem lacks room for a transfer.
+
+        Drivers call this from ``file_copy`` and ``remote_file_copy`` before starting a
+        transfer. The concrete per-platform probe is ``_get_free_space``; this helper
+        centralises the comparison, logging, and error shape so every driver behaves the
+        same way.
+
+        Args:
+            required_bytes (int): Number of bytes the pending transfer needs.
+            file_system (str, optional): Target filesystem passed through to
+                ``_get_free_space``. Drivers that auto-detect a filesystem should resolve
+                it before calling.
+
+        Raises:
+            NotEnoughFreeSpaceError: When the device reports fewer free bytes than
+                ``required_bytes``.
+        """
+        available = self._get_free_space(file_system)
+        log.debug(
+            "Host %s: filesystem %s has %s bytes free; %s bytes required.",
+            self.host,
+            file_system,
+            available,
+            required_bytes,
+        )
+        if available < required_bytes:
+            log.error(
+                "Host %s: insufficient free space on %s (%s free, %s required).",
+                self.host,
+                file_system,
+                available,
+                required_bytes,
+            )
+            raise NotEnoughFreeSpaceError(
+                hostname=self.host,
+                required=required_bytes,
+                available=available,
+                file_system=file_system,
+            )
+
+    def _get_free_space(self, file_system=None):
+        """Return the number of free bytes on ``file_system``.
+
+        Drivers override this to issue a platform-specific command (e.g., ``dir`` on
+        EOS/IOS/NXOS, ``show system storage`` on JUNOS) and parse the result. Drivers
+        whose platform exposes only one practical filesystem (e.g., EOS) may safely
+        ignore ``file_system``.
+
+        Args:
+            file_system (str, optional): The target filesystem to inspect. Drivers
+                that support multiple filesystems should treat ``None`` as "auto-detect
+                the default filesystem".
+
+        Returns:
+            int: Free bytes available on ``file_system``.
+
+        Raises:
+            NotImplementedError: When a driver has not yet implemented the probe.
+        """
+        raise NotImplementedError(f"{self.__class__.__name__} does not implement _get_free_space")
+
+    def _pre_flight_space_check(self, src, file_system=None):
+        """Run the free-space check iff ``src.file_size_bytes`` is populated.
+
+        Drivers call this from ``remote_file_copy`` so the check is skipped
+        (fail-open) when the caller omits ``FileCopyModel.file_size``; when set,
+        ``_check_free_space`` raises ``NotEnoughFreeSpaceError`` if the device
+        lacks room. Lives on ``BaseDevice`` so every driver inherits the same
+        shape without duplicating the if/else.
+
+        Args:
+            src (FileCopyModel): The source specification for the pending transfer.
+            file_system (str, optional): Target filesystem; passed through to
+                ``_check_free_space`` (and from there to ``_get_free_space``).
+        """
+        if src.file_size_bytes is not None:
+            self._check_free_space(src.file_size_bytes, file_system=file_system)
+        else:
+            log.debug(
+                "Host %s: no file_size on FileCopyModel; skipping pre-flight space check.",
+                self.host,
+            )
 
     def _image_booted(self, image_name, **vendor_specifics):
         """Determine if a particular image is serving as the active OS.
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -117,8 +117,8 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
         """
         raise NotImplementedError(f"{self.__class__.__name__} does not implement _get_free_space")
 
-    def _pre_flight_space_check(self, src, file_system=None):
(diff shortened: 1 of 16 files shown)


mattmiller87:
I am of the opinion that the file size is a required piece of information to add this feature. It is similar to file checksum as a required field.  To do the copy, you need to know certain things about the file.

To leave the file size optional, likely adds more complexity in the code to process and determine file size.

pyntc/devices/base_device.py
```diff
@@ -52,6 +55,90 @@ def __init__(self, host, username, password, device_type=None, **kwargs):  # noq
...
+        Returns:
+            int: Free bytes available on ``file_system``.
+
+        Raises:
+            NotImplementedError: When a driver has not yet implemented the probe.
+        """
+        raise NotImplementedError(f"{self.__class__.__name__} does not implement _get_free_space")
+
+    def _pre_flight_space_check(self, src, file_system=None):
+        """Run the free-space check iff ``src.file_size_bytes`` is populated.
```
mattmiller87:
```suggestion
        """Run the free-space check if ``src.file_size_bytes`` is populated.
```

pyntc/devices/base_device.py
```diff
@@ -52,6 +55,90 @@ def __init__(self, host, username, password, device_type=None, **kwargs):  # noq
...
+
+        Returns:
+            int: Free bytes available on ``file_system``.
+
+        Raises:
+            NotImplementedError: When a driver has not yet implemented the probe.
+        """
+        raise NotImplementedError(f"{self.__class__.__name__} does not implement _get_free_space")
+
+    def _pre_flight_space_check(self, src, file_system=None):
```
mattmiller87:
```suggestion
    def _pre_transfer_space_check(self, src, file_system=None):
```

pyntc/devices/eos_device.py
```diff
@@ -88,6 +88,33 @@ def _file_copy_instance(self, src, dest=None, file_system="/mnt/flash"):
         log.debug("Host %s: File copy instance %s.", self.host, file_copy)
         return file_copy
 
+    def _get_free_space(self, file_system=None):  # pylint: disable=unused-argument
+        """Return free bytes on ``file_system`` as reported by Arista's ``dir`` output.
+
+        EOS only exposes a single flash filesystem in practice, and ``dir`` always
+        prints ``NNNNN bytes total (MMMMM bytes free)`` as the last line, so the
```
mattmiller87: This reads a bit like device CLI documentation, which is fine.  I think I would prefer variable looking identifiers instead. So maybe we consider:
```suggestion
        prints ``<total> bytes total (<free> bytes free)`` as the last line, so the
```
or
```suggestion
        prints ``{total} bytes total ({free} bytes free)`` as the last line, so the
```

pyntc/devices/eos_device.py
```diff
@@ -88,6 +88,33 @@ def _file_copy_instance(self, src, dest=None, file_system="/mnt/flash"):
...
+
+        Returns:
+            int: Free bytes available on the target filesystem.
+
+        Raises:
+            CommandError: When the ``dir`` output does not contain a parseable
+                ``(N bytes free)`` trailer.
+        """
+        raw_data = self.show("dir", raw_text=True)
+        match = re.search(r"\((\d+)\s+bytes\s+free\)", raw_data)
```
mattmiller87, line 111:
```suggestion
       # Example: 3634421760 bytes total (1951289344 bytes free)
        match = re.search(r"\((\d+)\s+bytes\s+free\)", raw_data)
```

pyntc/devices/eos_device.py
```diff
@@ -606,6 +640,8 @@ def remote_file_copy(self, src: FileCopyModel, dest: str | None = None, file_sys
         self.open()
         self.enable()
 
+        self._pre_flight_space_check(src, file_system)
```
mattmiller87: If we change the above function name, we would need to update this
```suggestion
        self._pre_transfer_space_check(src, file_system)
```

pyntc/utils/models.py
```diff
@@ -7,6 +7,15 @@
 # Use Hashing algorithms from Nautobot's supported list.
 HASHING_ALGORITHMS = {"md5", "sha1", "sha224", "sha384", "sha256", "sha512", "sha3", "blake2", "blake3"}
 
+# Supported units for FileCopyModel.file_size, mapped to their multiplier in bytes.
+# Conversions use binary units (1 MB = 1024**2 bytes) to match network-device
+# filesystem reporting (e.g., Arista's ``dir`` output).
```
mattmiller87: I think this example is not needed here.
```suggestion
```

tests/integration/_helpers.py
mattmiller87, line 1: I like how this simplifies repeated logic in all the test_{device_type}_device.py files.

tests/integration/conftest.py
mattmiller87, line 1: This makes sense for simplifying and sharing the module for pytest.

tests/unit/test_devices/test_nxos_device.py
mattmiller87, line 1: We can do this, but NXOS will get a PR for the free-space check too. So we'll confirm this continues to work there.

tests/unit/test_devices/test_ios_device.py
mattmiller87, line 1: We can do this, but IOS will get a PR for the free-space check too. So we'll confirm this continues to work there.

tests/unit/test_devices/test_asa_device.py
mattmiller87, line 1: We can do this, but ASA will get a PR for the free-space check too. So we'll confirm this continues to work there.

tests/unit/test_devices/test_eos_device.py
mattmiller87, line 1: looks straight forward for the implementation.

tests/integration/test_asa_device.py
mattmiller87, line 1: We can do this, but ASA will get a PR for the free-space check too. So we'll confirm this continues to work there.

mattmiller87 (approved):
I have a few minor things I would change, which do not impact the scope or work accomplished here.

mattmiller87 (approved):
LGTM

jeffkala (approved):

PR #376, opened 2026-04-28
Add reboot flag to install_os

gsnider2195:
This required vendoring pynxos and adding the reboot flag support to that dependency. Long term plan is to retire the pynxos package and remove it from this project after migrating all nxos commands to Netmiko.


pyntc/devices/pynxos/features/vlans.py
```diff
@@ -0,0 +1,78 @@
...
+# class VlanNotInRangeError(NXOSError):
+#    def __init__(self):
+#        super(VlanNotInRangeError, self).__init__(
+#            'Vlan Id must be in range 1-3967')
+#
+# def vlan_not_in_range_error(vlan_id):
+#    vlan_id = int(vlan_id)
+#
+#    if vlan_id < 1 or vlan_id > 3967:
+#        raise VlanNotInRangeError
```
jeffkala: do we care that these are there?

pyntc/devices/pynxos/features/vlans.py
```diff
@@ -0,0 +1,78 @@
...
+
+    #    def get(self, vlan_id):
+    #        vlan_not_in_range_error(vlan_id)
+    #
+    #        vlan_id_table = self.device.show('show vlan id %s' % vlan_id)
+    #        try:
+    #            return converted_list_from_table(
+    #                vlan_id_table, 'vlanbriefid', VLAN_KEY_MAP)[0]
+    #        except IndexError:
+    #            return {}
```
jeffkala: same as previous

pyntc/devices/pynxos/features/vlans.py
```diff
@@ -0,0 +1,78 @@
...
+#        self.device.config_list(vlan_config_commands)
+#
+#    def remove(self, vlan_id):
+#        vlan_not_in_range_error(vlan_id)
+#
+#        vlan_remove_command = 'no vlan %s' % vlan_id
+#        self.device.config(vlan_remove_command)
+#
+# def instance(device):
+#    return Vlans(device)
```
jeffkala: same as previous

pyntc/devices/ios_device.py
```diff
@@ -926,6 +927,9 @@ def install_os(self, image_name, install_mode=False, read_timeout=2000, **vendor
                         self.reboot()
             else:
                 self.set_boot_options(image_name, **vendor_specifics)
+                if not reboot:
+                    log.info("Host %s: OS image %s boot options set. Reboot the device to apply", self.host, image_name)
+                    return True
```
jeffkala, line 937: Feel like for ios we need more logic.  If install_mode is used the reboot automatically happens.

gsnider2195: We could copy what we're doing in iosxewlc and fail if the user selects install mode and reboot=False

pyntc/devices/pynxos/features/vlans.py
gsnider2195: This code hasn't been touched in 7 years so I think it's safe to rip it out

jeffkala (approved):

PR #383, opened 2026-05-05
The pyntc rotating file handler is now opt-in

gsnider2195:
The pyntc rotating file handler is now opt-in via the `PYNTC_LOG_FILE ` environment variable.

### DIFF
--- a/pyntc/log.py
+++ b/pyntc/log.py
@@ -17,6 +17,10 @@ DEBUG_FORMAT = "%(asctime)s [%(levelname)s] [%(module)s] [%(funcName)s] %(name)s
 def get_log(name=None):
     """Get log namespace and creates logger and rotating file handler.
 
+    A :class:`RotatingFileHandler` is attached if the ``PYNTC_LOG_FILE``
+    environment variable is set, in which case its value is used as the log
+    file path.
+
     Args:
         name (str, optional): Sublogger name. Defaults to None.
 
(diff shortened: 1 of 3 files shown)


pyntc/log.py
```diff
@@ -17,17 +17,25 @@
...
-    handler = RotatingFileHandler(f"{logger_name}.log", maxBytes=2000)
     _logger = logging.getLogger(logger_name)
-    _logger.addHandler(handler)
+
+    log_file = os.environ.get("PYNTC_LOG_FILE")
+    if log_file and not any(
+        isinstance(h, RotatingFileHandler) and getattr(h, "baseFilename", None) == os.path.abspath(log_file)
+        for h in _logger.handlers
+    ):
+        _logger.addHandler(RotatingFileHandler(log_file, maxBytes=2000))
```
jeffkala, line 38: I might be reading this wrong, but this seems like a breaking change. So will we do a major release?

gsnider2195: Removing the logging to file as a breaking change? Is this a feature that people rely on in production or is it more of a development/testing feature?

jeffkala: 🤷 not sure what people use it for, but if folks are relying on it and we change to an opt-in that seems in theory like a breaking change.

jeffkala: happy to be overruled, just pointing it out as we attempt to follow semver

gsnider2195: I'm fine with a major version bump for this change

gsnider2195: `nornir-pyntc` has no upper bound so we won't need to update that and the other apps depend on `nornir-pyntc` so it shouldn't be too big of a deal

jeffkala (approved):

PR #395, opened 2026-05-27
NAPPS-968 | Adds device.install_mode to base driver and ios driver

jtdub:
# Closes NAPPS-968

## What's Changed:
- Adds `device.install_mode` property to base driver that will raise `NotImplemented`.
- Adds `device.install_mode` property to IOS driver that will return `True` if `device.boot_options.get("sys") == "packages.conf"`. Otherwise it will return `False`.
- Deprecates the `install_mode` argument in the IOS driver.


pyntc/devices/ios_device.py
```diff
@@ -877,13 +889,29 @@ def file_copy_remote_exists(self, src, dest=None, file_system=None):
...
+    def _resolve_install_mode(self, install_mode):
+        """Return the effective install_mode flag, warning if the caller passed it explicitly."""
+        if install_mode is None:
+            return self.install_mode
+        warnings.warn(
+            "The install_mode argument to install_os is deprecated; install mode is now "
+            "derived from the device's boot_options via the install_mode property.",
+            DeprecationWarning,
+            stacklevel=3,
+        )
```
Pkawa-ntc, line 900: Nitpick, but is that `stacklevel=3` intentional? I see that we use .warn() in other places with default stacklevel=2 for deprecation

changes/395.added
```diff
@@ -0,0 +1 @@
+Added an ``install_mode`` property to ``BaseDevice`` (abstract) and ``IOSDevice``; the IOS implementation returns ``True`` when the device boots from ``packages.conf``.
```
Pkawa-ntc, line 1: if it add things, shouldn't it be `395.added`?

Pkawa-ntc (approved):
Some nits to consider, other than that LGTM. Thanks James!

PR #407, opened 2026-07-21
Fixed Arista EOS reboots not being detected when waiting for the device to reload

Defiantearth:
### What's Changed
- Fixed Arista EOS reboots not being detected when waiting for the device to reload.

Bug in OS upgrades found where when upgrading a device pyntc would return in `_wait_for_device_reboot()` before the device actually rebooted.

### Note: You can see `Checking for reachability`. This is OS Upgrades taking over waiting for the device to come back up since pyntc returned before the device came back up.
localhost_8080_extras_job-results_6d38ccfc-a372-47f8-bada-a7ae2ea2e9bf__page=2

### Aft
localhost_8080_extras_job-results_8ae663bb-d9df-4b88-a7df-7b54660a2784_ (1)
er

### DIFF
--- a/pyntc/devices/eos_device.py
+++ b/pyntc/devices/eos_device.py
@@ -178,34 +178,36 @@ class EOSDevice(BaseDevice):
 
         return f"{days:02d}:{hours:02d}:{mins:02d}:{seconds:02d}"
 
-    def _wait_for_device_reboot(self, original_uptime, timeout=DEFAULT_REBOOT_TIMEOUT):
-        """Block until the device reboots, detected by uptime dropping below original_uptime.
+    def _wait_for_device_reboot(self, original_boot_time, timeout=DEFAULT_REBOOT_TIMEOUT):
+        """Block until the device reboots, detected by its boot time advancing past original_boot_time.
 
         Args:
-            original_uptime (int): Device uptime in seconds captured before the reboot.
+            original_boot_time (float): Device boot time (epoch seconds) captured before the reboot.
             timeout (int): Max seconds to poll for the device to return. Defaults to 3600.
 
         Raises:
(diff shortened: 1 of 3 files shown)


pyntc/devices/eos_device.py
```diff
@@ -177,15 +178,40 @@ def _uptime_to_string(self, uptime):
 
         return f"{days:02d}:{hours:02d}:{mins:02d}:{seconds:02d}"
 
-    def _wait_for_device_reboot(self, timeout=3600):
+    def _wait_for_device_reboot(self, original_boot_time, timeout=DEFAULT_REBOOT_TIMEOUT):
```
Defiantearth, line 181: Using boot time instead of uptime here for the case where a devices has a uptime of 30 sec upgrades then takes a few min such that the uptime is greater then 30 sec when the api is ready which will result in a timeout.

pyntc/devices/eos_device.py
```diff
@@ -729,14 +764,13 @@ def install_os(self, image_name, reboot=True, **vendor_specifics):
...
         """
-        timeout = vendor_specifics.get("timeout", 3600)
+        timeout = vendor_specifics.get("timeout", DEFAULT_REBOOT_TIMEOUT)
         if not self._image_booted(image_name):
             self.set_boot_options(image_name, **vendor_specifics)
             if not reboot:
                 log.info("Host %s: OS image %s boot options set. Reboot the device to apply", self.host, image_name)
                 return True
-            self.reboot()
-            self._wait_for_device_reboot(timeout=timeout)
```
Defiantearth, line 739: This was already in reboot() So as long as we pass in the timeout and wait_for_reload=True we can remove the call here.

jtdub (approved):
LGTM!

PR #411, opened 2026-07-31
Fix checksum verification timing out on large OS images

### LINKED ISSUE #410: Verify /md5 times out on large OS images in file_copy and remote_file_copy
### Environment
* Python version: 3.10
* pyntc version: 3.2.1 

### Expected Behavior
After transferring a large OS image (e.g. a ~1.2 GB Cat9k image), the checksum
verification step should be given enough time to complete.

### Observed Behavior
[`get_remote_checksum`](https://github.com/networktocode/pyntc/blob/1bdb5a03f41fda18674fb8fce18ffac7be9b96bc/pyntc/devices/ios_device.py#L666) hardcodes the timeout for `verify /md5`:

    result = self.native.send_command_timing(cmd, read_timeout=300)

`file_copy` and `remote_file_copy`
both call `verify_file -> compare_file_checksum -> get_remote_checksum` before
and after the transfer — and no caller can pass a different value; nothing in
the chain accepts a timeout.

On Catalyst 9k hardware, `verify /md5` of a ~1.2 GB image takes longer than
300 seconds while continuously printing progress, so netmiko raises:

    read_channel_timing's absolute timer expired.
    The network device was continually outputting data for longer than 300 seconds.

The result is a transfer that completed successfully on the device but is
reported as failed. Since the observed hashing rate is under 4 MB/s, current
2+ GB images need 500-700+ seconds. Retried jobs can also fail in the
*pre*-transfer verification, since the file already exists on flash by then.

The same hardcode exists in `asa_device.py:677` and `iosxr_device.py:437`.

Related defect in the same function: in `get_remote_checksum` the
`return match[1]` (ios_device.py:678) sits outside the `if match := ...` block,
inside the `for` loop. When the first pattern does not match — which is exactly
what happens when the output is truncated by this timeout — it raises
`TypeError: 'NoneType' object is not subscriptable` instead of trying the second
pattern or raising the intended `CommandError`. 

### Steps to Reproduce
1. Pick an image large enough that `verify /md5` takes more than 300 seconds on
   the target device (any recent Cat9k image on C9200/C9300 works).
2. Transfer it with `device.file_copy(...)` or `device.remote_file_copy(...)`
   against a Cisco IOS-XE device, without `read_timeout_override` set on the
   connection.
3. The transfer itself succeeds, then the post-transfer verification dies at
   300 seconds with the netmiko `ReadTimeout` above (surfacing as the
   `TypeError` from the related defect).

michalis1:
Fixes #410

## Problem

[`get_remote_checksum`](https://github.com/networktocode/pyntc/blob/1bdb5a03f41fda18674fb8fce18ffac7be9b96bc/pyntc/devices/ios_device.py#L666) hardcodes `read_timeout=300` for the checksum command
(`verify /md5` on IOS/ASA, `run md5sum` on IOS-XR), and nothing in the
`verify_file -> compare_file_checksum -> get_remote_checksum` chain accepts a
timeout. On Catalyst 9k hardware `verify /md5` hashes at under 4 MB/s, so a
~1.2 GB image already exceeds 300s and current 2+ GB images need 500-700+
seconds. The result: `file_copy` / `remote_file_copy` transfers that completed
successfully on the device get reported as failures when the post-transfer
verification hits netmiko's absolute read timer. Retried jobs can also fail in
the pre-transfer verification, since the file is already on flash by then.

### DIFF
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -410,6 +410,8 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
             file_system (str): Supported only for IOS and NXOS. The file system for the
                 remote file. If no file_system is provided, then the ``get_file_system``
                 method is used to determine the correct file system to use.
+            read_timeout (int): Supported only for IOS, ASA and IOS-XR. Maximum time in
+                seconds to wait for the checksum command to complete (default: 900).
 
         Returns:
             (bool): True if the checksums match, False otherwise.
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -90,7 +90,7 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
                 required_bytes,
             )
             raise NotEnoughFreeSpaceError(
-                self.host,
+                hostname=self.host,
                 required=required_bytes,
                 available=available,
                 file_system=file_system,
--- a/pyntc/devices/base_device.py
+++ b/pyntc/devices/base_device.py
@@ -90,7 +90,7 @@ class BaseDevice:  # pylint: disable=too-many-instance-attributes,too-many-publi
                 required_bytes,
             )
             raise NotEnoughFreeSpaceError(
-                hostname=self.host,
+                self.host,
                 required=required_bytes,
                 available=available,
                 file_system=file_system,
--- a/pyntc/devices/ios_device.py
+++ b/pyntc/devices/ios_device.py
@@ -140,7 +140,7 @@ class IOSDevice(BaseDevice):
                 continue
 
         log.error("host %s: File system not found with command 'dir'.")
-        raise FileSystemNotFoundError(hostname=self.hostname, command="dir")
+        raise FileSystemNotFoundError(self.hostname, "dir")
 
     def _get_free_space(self, file_system=None):
         """Return free bytes on ``file_system`` as reported by IOS ``dir`` output."""
@@ -263,7 +263,7 @@ class IOSDevice(BaseDevice):
                 time.sleep(10)
 
         log.error("Host %s: Device timed out while rebooting.", self.host)
-        raise RebootTimeoutError(hostname=self.hostname, wait_time=timeout)
+        raise RebootTimeoutError(self.hostname, timeout)
 
     def _has_reload_happened_recently(self):
         if re.search(r"^00:00:0\d:*", self.uptime_string) is None:
@@ -904,7 +904,9 @@ class IOSDevice(BaseDevice):
         )
         return install_mode
 
-    def install_os(self, image_name, reboot=True, install_mode=None, read_timeout=2000, **vendor_specifics):
+    def install_os(  # pylint: disable=too-many-branches
+        self, image_name, reboot=True, install_mode=None, read_timeout=2000, **vendor_specifics
+    ):
         """Installs the prescribed Network OS, which must be present before issuing this command.
 
         Args:
@@ -955,12 +957,25 @@ class IOSDevice(BaseDevice):
                         install_message = self.show(command, read_timeout=read_timeout)
                         if install_message.startswith("FAILED:"):
                             log.error("Host %s: OS install error for image %s", self.host, image_name)
-                            raise OSInstallError(hostname=self.hostname, desired_boot=image_name)
+                            raise OSInstallError(self.hostname, image_name, install_message)
                     except IOError:
                         log.error("Host %s: IO error for image %s", self.host, image_name)
-                    except CommandError:
+                    except CommandError as original_error:
+                        log.warning(
+                            "Host %s: install command failed (%s); falling back to legacy "
+                            "'request platform software package install' command.",
+                            self.host,
+                            original_error.cli_error_msg,
+                        )
                         command = f"request platform software package install switch all file {self._get_file_system()}{image_name} auto-copy"
-                        self.show(command, read_timeout=read_timeout)
+                        try:
+                            self.show(command, read_timeout=read_timeout)
+                        except CommandError as fallback_error:
+                            raise CommandError(
+                                command,
+                                f"{fallback_error.cli_error_msg} (legacy fallback; original install "
+                                f"error: {original_error.cli_error_msg})",
+                            ) from original_error
                         self.reboot()
             else:
                 self.set_boot_options(image_name, **vendor_specifics)
@@ -979,7 +994,7 @@ class IOSDevice(BaseDevice):
             # Verify the OS level
             if not self._image_booted(image_name):
                 log.error("Host %s: OS install error for image %s", self.host, image_name)
-                raise OSInstallError(hostname=self.hostname, desired_boot=image_name)
+                raise OSInstallError(self.hostname, image_name)
 
             log.info("Host %s: OS image %s installed successfully.", self.host, image_name)
             return True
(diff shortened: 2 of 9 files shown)


pyntc/devices/ios_device.py
```diff
@@ -675,7 +677,7 @@ def get_remote_checksum(self, filename, hashing_algorithm="md5", file_system=Non
                     hashing_algorithm,
                     match[1],
                 )
-            return match[1]
+                return match[1]
```
jeffkala, line 680: extra indent expected?

pyntc/devices/iosxr_device.py
```diff
@@ -481,7 +483,7 @@ def check_file_exists(self, filename, file_system=None):
         log.debug("Host %s: File %s not found in 'dir' output on %s.", self.host, filename, file_system)
         return False
 
-    def verify_file(self, checksum, filename, hashing_algorithm="md5", file_system=None):
+    def verify_file(self, checksum, filename, hashing_algorithm="md5", file_system=None, read_timeout=900):
```
jeffkala, line 486: Is 900 a good sane default?

jeffkala (commented):
Looks good to me, I'd like @jtdub or @gsnider2195 to review as well since they've been so in depth with this lib lately.

pyntc/devices/iosxr_device.py
michalis1: The old 300s default can't verify current images: we measured `verify /md5` on Cat9200 hardware hashing at under 4 MB/s, so a ~1.2 GB image already exceeds 300s, and today's 2+ GB images need 500–700s+. 600 would be marginal; 900 covers them with headroom and matches the existing `FileCopyModel.timeout` default.

pyntc/devices/ios_device.py
michalis1: The `return match[1]` was sitting outside the `if match := re.search(...)` block but inside the for loop, so it executed unconditionally on the first iteration. When the first pattern didn't match — which is exactly what happens when the verify output is truncated by a timeout — `match` was None and this raised `TypeError: 'NoneType' object is not subscriptable`. The second pattern was never tried, and the intended `CommandError` below the loop was unreachable.

Moving the `return` inside the `if` restores the obvious intent: try each pattern, return on the first match, and fall through to `CommandError` if none match. No behavior change on the happy path — pattern 1 matching still returns immediately.

pyntc/devices/ios_device.py
```diff
@@ -951,12 +957,25 @@ def install_os(self, image_name, reboot=True, install_mode=None, read_timeout=20
                         install_message = self.show(command, read_timeout=read_timeout)
                         if install_message.startswith("FAILED:"):
                             log.error("Host %s: OS install error for image %s", self.host, image_name)
-                            raise OSInstallError(hostname=self.hostname, desired_boot=image_name)
+                            raise OSInstallError(self.hostname, image_name, install_message)
                     except IOError:
                         log.error("Host %s: IO error for image %s", self.host, image_name)
-                    except CommandError:
+                    except CommandError as original_error:
```
jtdub, line 965: Can we add some testing around this exception?

pyntc/devices/ios_device.py
```diff
@@ -951,12 +957,25 @@ def install_os(self, image_name, reboot=True, install_mode=None, read_timeout=20
...
+                            "Host %s: install command failed (%s); falling back to legacy "
+                            "'request platform software package install' command.",
+                            self.host,
+                            original_error.cli_error_msg,
+                        )
                         command = f"request platform software package install switch all file {self._get_file_system()}{image_name} auto-copy"
-                        self.show(command, read_timeout=read_timeout)
+                        try:
+                            self.show(command, read_timeout=read_timeout)
+                        except CommandError as fallback_error:
```
jtdub, line 975: Can we add some testing around this exception?

pyntc/devices/base_device.py
```diff
@@ -90,7 +90,7 @@ def _check_free_space(self, required_bytes, file_system=None):
                 required_bytes,
             )
             raise NotEnoughFreeSpaceError(
-                hostname=self.host,
```
jtdub, line 93: We should revert the places were we are omitting keyword arguments.

pyntc/devices/ios_device.py
```diff
@@ -140,7 +140,7 @@ def _get_file_system(self):
                 continue
 
         log.error("host %s: File system not found with command 'dir'.")
-        raise FileSystemNotFoundError(hostname=self.hostname, command="dir")
```
jtdub, line 143: We should revert the places were we are omitting keyword arguments.

pyntc/devices/ios_device.py
```diff
@@ -263,7 +263,7 @@ def _wait_for_device_reboot(self, timeout=3600):
                 time.sleep(10)
 
         log.error("Host %s: Device timed out while rebooting.", self.host)
-        raise RebootTimeoutError(hostname=self.hostname, wait_time=timeout)
```
jtdub, line 266: We should revert the places were we are omitting keyword arguments.

pyntc/devices/ios_device.py
```diff
@@ -951,12 +957,25 @@ def install_os(self, image_name, reboot=True, install_mode=None, read_timeout=20
                         install_message = self.show(command, read_timeout=read_timeout)
                         if install_message.startswith("FAILED:"):
                             log.error("Host %s: OS install error for image %s", self.host, image_name)
-                            raise OSInstallError(hostname=self.hostname, desired_boot=image_name)
```
jtdub, line 954: We should revert the places were we are omitting keyword arguments.

pyntc/devices/ios_device.py
```diff
@@ -975,7 +994,7 @@ def install_os(self, image_name, reboot=True, install_mode=None, read_timeout=20
             # Verify the OS level
             if not self._image_booted(image_name):
                 log.error("Host %s: OS install error for image %s", self.host, image_name)
-                raise OSInstallError(hostname=self.hostname, desired_boot=image_name)
```
jtdub, line 978: We should revert the places were we are omitting keyword arguments.

pyntc/devices/ios_device.py
```diff
@@ -1234,7 +1253,7 @@ def set_boot_options(self, image_name, **vendor_specifics):
         file_system_files = self.show(f"dir {file_system}")
         if image_name != INSTALL_MODE_FILE_NAME and re.search(image_name, file_system_files) is None:
             log.error("Host %s: File not found error for image %s.", self.host, image_name)
-            raise NTCFileNotFoundError(hostname=self.hostname, file=image_name, directory=file_system)
```
jtdub, line 1237: We should revert the places were we are omitting keyword arguments.

jtdub (changes requested):

pyntc/devices/ios_device.py
michalis1: done

pyntc/devices/ios_device.py
michalis1: done

pyntc/devices/ios_device.py
michalis1: done

pyntc/devices/ios_device.py
michalis1: done

pyntc/devices/ios_device.py
michalis1: done

pyntc/devices/base_device.py
michalis1: done

pyntc/devices/ios_device.py
michalis1: done

pyntc/devices/ios_device.py
michalis1: done

jtdub (approved):

PR #419, opened 2026-08-11
Add arista_eos_ssh driver for Arista EOS over SSH

jvanderaa:
Adds EOSSSHDevice for environments where eAPI is not enabled. It subclasses EOSDevice and overrides only the transport surface (__init__, open, close, show, config, reboot, vlans) plus a native_ssh alias, so the fact properties, boot-option handling, install_os and the whole file-transfer family are inherited unchanged.

Structured data comes from the CLI's '| json' pipe, which renders the same document eAPI returns; the pipe is a CLI feature and does not require 'management api http-commands'. The pipe is gated on '^show' because EOSDevice routes five EXEC/config commands through show(raw_text=False), and per-command read timeouts are resolved from the command text so show()'s signature stays identical to EOSDevice.show().

Validated against a DCS-7050TX-64-R running EOS 4.28.5M; the unit fixtures are that capture, sanitised. Public API parity with EOSDevice is enforced by test rather than by convention, and a further test feeds both drivers the same document and requires identical facts.

### DIFF
--- a/changes/418.added
+++ b/changes/418.added
@@ -1,0 +1,1 @@
+Added the `arista_eos_ssh` device type, an SSH-only Arista EOS driver for environments where eAPI is not enabled; it exposes the same API as `arista_eos_eapi` and obtains structured data via the CLI's `| json` pipe.
--- a/docs/user/lib_getting_started.md
+++ b/docs/user/lib_getting_started.md
@@ -11,15 +11,18 @@ The first way is to use the `ntc_device` object. Just pass in all required param
 
 Like many libraries, we need to pass in the host/IP and credentials.  Because this is a multi-vendor/API library, we also use the `device_type` parameter to identify which device we are building an instance of.
 
-pyntc currently supports seven device types:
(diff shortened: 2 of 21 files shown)


jeffkala:
Ran some additional manual testing on real hardware as well.

1. Connect
2. Facts
3. show() variants
4. Error handling
5. Configs
6. Filesystem
7. Config round trip (state change, self-reverting)
8. Checkpoint / rollback (state change, self-reverting)
9. Save (PERSISTS: copies running-config to startup-config)
10. Close / reconnect

jtdub:
@jvanderaa - Does this need to be supported by os upgrades?

jtdub (approved):
LGTM! I ran the integration tests against the device in the lab. 

I'm in the process of creating a Jira issue to wrap up anything else that would allow us to use this driver in OS Upgrades.

jeffkala:
> @jvanderaa - Does this need to be supported by os upgrades?

yes

PR #427, opened 2026-08-17
Keep the byte counts on NotEnoughFreeSpaceError

### LINKED ISSUE #426: NotEnoughFreeSpaceError discards its byte counts, forcing callers to parse the message
### Environment
* pyntc version: 3.2.3a1 (develop)

### Observed Behavior

`NotEnoughFreeSpaceError` accepts `required`, `available` and `file_system`, uses them to build the message string, and then calls `super().__init__(message)` without storing any of them ([errors.py L209-L226](https://github.com/networktocode/pyntc/blob/develop/pyntc/errors.py#L209-L226)). The values are only recoverable by parsing `message`.

The message also prints the raw integers with no thousands separators, so a nine-digit byte count is hard to read and the shortfall has to be worked out by hand:

```
host1: bootflash: has 13456789 bytes free; 313456789 bytes required for transfer
```

### Expected Behavior

The byte counts kept as attributes, so a caller can render its own message or compute the shortfall without a regex. The default message readable at a glance, including how much more space is needed.

### Steps to Reproduce

```python
>>> from pyntc.errors import NotEnoughFreeSpaceError
>>> err = NotEnoughFreeSpaceError(hostname="host1", required=313456789, available=13456789, file_system="bootflash:")
>>> err.required
AttributeError: 'NotEnoughFreeSpaceError' object has no attribute 'required'
```

Found while improving disk-space log messages in a downstream app, which currently has to regex the values back out of `message`.

bernalvinicius:
Closes #426

## Explanation of Change(s)

`NotEnoughFreeSpaceError` accepted `required`, `available` and `file_system`, used them to build the message string, and then dropped them — the values were only recoverable by parsing `message`. They are now kept as attributes, along with a precomputed `shortfall`.

The message also gained thousands separators and the shortfall, since a nine-digit byte count is hard to read and the remaining space was left for the reader to work out:

```
- host1: bootflash: has 13456789 bytes free; 313456789 bytes required for transfer
+ host1: bootflash: has 13,456,789 bytes free; 313,456,789 bytes required for transfer (300,000,000 more bytes required to succeed)
```

The legacy `min_space` form is untouched; its byte attributes are `None`.

**Heads-up on the message change:** the wording is observable to anyone parsing it. That is exactly the situation this fixes — [nautobot-app-os-upgrades](https://github.com/networktocode-llc/nautobot-app-os-upgrades) was regexing the counts back out of `message` and drops that workaround once this lands.

## Added change log fragment(s)

`changes/426.changed`

## Unit, Integration Tests

Four new tests in `tests/unit/test_errors.py` cover the stored attributes, the message with counts and shortfall, the message without a file system, and the legacy form leaving the byte attributes unset. Full suite: 950 passed, 26 skipped.

## Outline Remaining Work, Constraints from Design

None.

### DIFF
--- a/pyntc/errors.py
+++ b/pyntc/errors.py
@@ -210,6 +210,10 @@ class NotEnoughFreeSpaceError(NTCError):
         """
         Error for not having enough free space to transfer a file.
 
+        The byte counts are kept as attributes so callers can render their own message
+        (or compute the shortfall) without parsing ``message``. They are ``None`` when the
+        error is raised in the legacy ``min_space`` form.
+
         Args:
             hostname (str): The hostname of the device being checked.
             min_space (str, optional): The minimum amount of space required. Retained for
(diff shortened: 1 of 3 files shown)


gsnider2195 (approved):
Looks great thanks!

pyntc/errors.py
```diff
@@ -210,17 +210,39 @@ def __init__(self, hostname, min_space=None, *, required=None, available=None, f
         """
         Error for not having enough free space to transfer a file.
 
+        The byte counts are kept as attributes so callers can render their own message
+        (or compute the shortfall) without parsing ``message``. They are ``None`` when the
+        error is raised in the legacy ``min_space`` form.
```
jvanderaa:
```suggestion
        (or compute the shortfall) without parsing `message`. They are `None` when the
        error is raised in the legacy `min_space` form.
```

jvanderaa (commented):
Update to single backticks.

PR #429, opened 2026-09-17
Fix IOS remote file copy credentials and report device copy errors

mattmiller87:
## Summary

FTP transfers to Cisco IOS devices never authenticated. `FileCopyModel.__post_init__` strips credentials out of the download URL (`pyntc/utils/models.py:88-90`), and the IOS driver sent that stripped URL to the device. It relied on answering `Source username` and `Password` prompts instead. IOS does not issue those prompts for `copy ftp:`. The device attempted an anonymous login, the server refused it, and the transfer failed in under half a second.

The failure was reported through a Nautobot OS upgrade workflow, which surfaced this:

```
FileTransferError:
An error occurred during transfer. Please make sure the local file exists and
that appropriate permissions are set on the remote device.
```

## What Changed

Two commits, because these are two separate defects:

1. **`Fix IOS remote file copy dropping URL credentials`** is the reported bug.
2. **`Report the device error when a remote file copy fails`** improved logging across four drivers.

### Changes by driver

Repurposed existing methods to make all drivers consistent

| Element | Copied from |
| --- | --- |
| `_netloc`, `_source_path`, `_build_url_copy_command_*` | `nxos_device.py:413-438` |
| `for/else` guard on the prompt loop | `asa_device.py:1110-1123` |
| Error carrying the device output | `eos_device.py:633-637` |
| Token masking | `jnpr_device.py:1829` |
| `or ""` on the prompt answers | `nxos_device.py:596-597` |

**`ios_device.py`** gets the credential routing, the prompt-answer defaults, the error detail and the loop guard.

**`nxos_device.py`**, **`asa_device.py`** and **`iosxr_device.py`** get the error detail only, plus the loop guard where they lacked one. ASA and IOS-XR already logged the device output and then threw it away at the exception.

## Testing

The driver was run end to end against a live device, a WS-C3650-48PQ running IOS-XE 16.12.14. 

```
INFO pyntc: Host [redacted]: File pyntc_ftp_test.bin transferred successfully with output: Accessing ftp://*:*@[redacted]/IOS-XE/pyntc_ftp_test.bin...!
[OK - 94038/4096 bytes]

94038 bytes copied in 0.159 secs (591434 bytes/sec)

PASSED: flash:pyntc_ftp_test.bin transferred and the md5 verified.
```

Fifteen new unit tests cover the IOS credential routing, the non-default port, the VRF token, the error detail and the loop guard. The error and loop-guard tests are mirrored on NX-OS, ASA and IOS-XR.

`tests/integration/test_ios_device.py` is new and skips unless its URL environment variables are set, so it runs against a lab device on demand and never in CI.

## Follow-ups, not in this PR

- `FileCopyModel.ftp_passive` does nothing outside the Junos driver. IOS defaults to active FTP and would need `ip ftp passive` in global config.
- Six drivers repeat the same copy-command construction and prompt walk. The logic could be collapsed into a utils.py or base.py helper. That should likely be a separate PR, if desired.


jtdub:
**PRAISE**: The PR fixes the root cause. IOS does not prompt for FTP credentials, so the driver now puts them into the URL. `HideUrlCredentials` applies only to the schemes that carry credentials. The comment at `ios_device.py:949` records why the log messages use f-strings. The error detail and the `for/else` loop guards have the same tests on IOS, NX-OS, ASA, and IOS-XR.

pyntc/devices/ios_device.py
```diff
@@ -795,7 +855,46 @@ def file_copy(self, src, dest=None, file_system=None):
...
+    def _build_url_copy_command_with_creds(self, src: FileCopyModel, file_system: str, dest: str) -> str:
+        """Build the copy command for transfers where IOS reads the credentials from the URL.
+
+        FTP, HTTP and HTTPS never prompt. A URL without credentials makes the device
+        attempt an anonymous login, which the server rejects.
+        """
+        netloc = self._netloc(src)
+        path = self._source_path(src, dest)
+        credentials = f"{src.username}:{src.token}" if src.token else src.username
+        return f"copy {src.scheme}://{credentials}@{netloc}{path} {file_system}{dest}"
```
jtdub: **MUST FIX**: This URL uses `src.path`, so the query string is lost. The old command used `src.clean_url`, which keeps `?query` (`pyntc/utils/models.py:92-93`). An FTP, HTTP, or HTTPS URL with a username and a query string now downloads the wrong resource. Add the query string from `urlparse(src.download_url).query` back to the URL. Also add a unit test for a URL with a query string.

pyntc/devices/ios_device.py
```diff
@@ -36,6 +37,65 @@
...
+
+    def __init__(self, native, token):
+        """Capture the registries the token is added to and removed from.
+
+        Args:
+            native (BaseConnection): The netmiko connection whose no_log mapping the token joins.
+            token (str): The password sent in the copy URL. A falsy value is a no-op.
+        """
+        self.token = token
+        self.netmiko_no_log = native._secrets_filter.no_log  # pylint: disable=protected-access
```
jtdub: **MUST FIX**: `native._secrets_filter` does not exist in netmiko 4.0.0 to 4.3.0. The attribute is present from 4.4.0. `pyproject.toml` allows `netmiko = ">=4.0,<5.0"`, so each IOS FTP, HTTP, or HTTPS copy on those versions raises `AttributeError`. Set the netmiko lower limit to `>=4.4`, or read the attribute with `getattr` and a fallback that keeps the pyntc filter.

pyntc/devices/ios_device.py
```diff
@@ -795,7 +855,46 @@ def file_copy(self, src, dest=None, file_system=None):
...
+
+    def _build_url_copy_command_with_creds(self, src: FileCopyModel, file_system: str, dest: str) -> str:
+        """Build the copy command for transfers where IOS reads the credentials from the URL.
+
+        FTP, HTTP and HTTPS never prompt. A URL without credentials makes the device
+        attempt an anonymous login, which the server rejects.
+        """
+        netloc = self._netloc(src)
+        path = self._source_path(src, dest)
+        credentials = f"{src.username}:{src.token}" if src.token else src.username
```
jtdub: **MUST FIX**: The code does not percent-encode `src.username` and `src.token` before it puts them in the URL. If the caller supplies a password with `@`, `/`, `:`, `?`, or `#` as an argument, IOS reads the wrong host or path. An example is `token="p@ss/w0rd"`. Encode both values with `urllib.parse.quote(..., safe="")`. Then mask the encoded token in `_mask_token` and in `HideUrlCredentials` as well. Credentials that come from `download_url` are already encoded, so do not encode them a second time.

